PR #3, opened 2015-02-05
Create S3SplitFileInput for reading data out of S3.

mreid-moz:
See [Bug 1126471](https://bugzilla.mozilla.org/show_bug.cgi?id=1126471).
This also adds the "heka-s3list" and "heka-s3cat" commands which can be used together or separately to look at the data in S3.

### DIFF
--- a/heka/cmd/heka-s3cat/main.go
+++ b/heka/cmd/heka-s3cat/main.go
@@ -1,0 +1,149 @@
+/***** BEGIN LICENSE BLOCK *****
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# The Initial Developer of the Original Code is the Mozilla Foundation.
+# Portions created by the Initial Developer are Copyright (C) 2014
+# the Initial Developer. All Rights Reserved.
+#
+# Contributor(s):
+#   Mark Reid (mreid@mozilla.com)
+# ***** END LICENSE BLOCK *****/
+
+/*
+
+A command-line utility for counting, viewing, filtering, and extracting Heka
+protobuf logs from files on Amazon S3.
+
+*/
+package main
+
+import (
+	"bufio"
+	"code.google.com/p/gogoprotobuf/proto"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"github.com/mozilla-services/heka/message"
+	"io"
+	"os"
+	"time"
+	"github.com/crowdmob/goamz/aws"
+	"github.com/crowdmob/goamz/s3"
+	"github.com/mreid-moz/s3splitfile"
+)
+
+func main() {
+	flagMatch := flag.String("match", "TRUE", "message_matcher filter expression")
+	flagFormat := flag.String("format", "txt", "output format [txt|json|heka|count]")
+	flagOutput := flag.String("output", "", "output filename, defaults to stdout")
+	flagStdin := flag.Bool("stdin", false, "read list of s3 key names from stdin")
+    flagBucket := flag.String("bucket", "default-bucket", "S3 Bucket name")
+    flagAWSKey := flag.String("aws-key", "DUMMY", "AWS Key")
+    flagAWSSecretKey := flag.String("aws-secret-key", "DUMMY", "AWS Secret Key")
+    flagAWSRegion := flag.String("aws-region", "us-west-2", "AWS Region")
+	flag.Parse()
+
+	if !*flagStdin && flag.NArg() < 1 {
+		flag.PrintDefaults()
+		os.Exit(1)
+	}
+
+	var err error
+	var match *message.MatcherSpecification
+	if match, err = message.CreateMatcherSpecification(*flagMatch); err != nil {
+		fmt.Printf("Match specification - %s\n", err)
+		os.Exit(2)
+	}
+
+	var out *os.File
+	if "" == *flagOutput {
+		out = os.Stdout
+	} else {
+		if out, err = os.OpenFile(*flagOutput, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644); err != nil {
+			fmt.Printf("%s\n", err)
+			os.Exit(3)
+		}
+		defer out.Close()
+	}
+
+	auth := aws.Auth{AccessKey: *flagAWSKey, SecretKey: *flagAWSSecretKey}
+	region, ok := aws.Regions[*flagAWSRegion]
+	if !ok {
+		fmt.Printf("Parameter 'aws-region' must be a valid AWS Region\n")
+		os.Exit(4)
+	}
+	s := s3.New(auth, region)
+	bucket := s.Bucket(*flagBucket)
+
+	if *flagStdin {
+		scanner := bufio.NewScanner(os.Stdin)
+	    for scanner.Scan() {
+	        filename := scanner.Text()
+			cat(bucket, filename, match, *flagFormat, out)
+	    }
+	} else {
+		for _, filename := range flag.Args() {
+			cat(bucket, filename, match, *flagFormat, out)
+		}
+	}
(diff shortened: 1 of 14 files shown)


heka/plugins/s3splitfile/s3splitfile_input.go
```diff
@@ -0,0 +1,203 @@
...
+	runner.LogMessage(fmt.Sprintf("Preparing to read: %s", s3Key))
+
+	if input.bucket == nil {
+		runner.LogMessage(fmt.Sprintf("Dude, where's my bucket: %s", s3Key))
+		return
+	}
+
+	deliverer := runner.NewDeliverer(s3Key)
+	defer deliverer.Done()
+	sr := runner.NewSplitterRunner(s3Key)
```
rafrombrc: You can create these in `fetcher` and pass them in to `readS3File` to reuse them. Saves us from spinning goroutines up and down each time.

heka/cmd/heka-s3cat/main.go
```diff
@@ -0,0 +1,149 @@
...
+		n := r.BytesRead
+		record := r.Record
+		err := r.Err
+
+		if err != nil && err != io.EOF {
+			fmt.Printf("Error reading %s: %s\n", s3Key, err)
+		} else {
+			if len(record) > 0 {
+				processed += 1
+				headerLen := int(record[1]) + message.HEADER_FRAMING_SIZE
```
trink, line 100: The fact that you have to decompose the framing yourself is something we should probably address in the Heka library

heka/cmd/heka-s3cat/main.go
```diff
@@ -0,0 +1,149 @@
+/***** BEGIN LICENSE BLOCK *****
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# The Initial Developer of the Original Code is the Mozilla Foundation.
+# Portions created by the Initial Developer are Copyright (C) 2014
```
trink: We should really not have copyright dates or contributors listed in individual source files (they just end up wrong like above). @rafrombrc

heka/cmd/heka-s3list/main.go
```diff
@@ -0,0 +1,98 @@
...
+	var b *s3.Bucket
+
+	prefix := strings.Trim(*flagBucketPrefix, "/")
+	if prefix != "" {
+		prefix += "/"
+	}
+
+	// Initialize the S3 bucket
+	auth := aws.Auth{AccessKey: *flagAWSKey, SecretKey: *flagAWSSecretKey}
+	s := s3.New(auth, aws.Regions[*flagAWSRegion])
```
trink: Why the different handling of region (cat vs list)

heka/patches/0003-Add-more-cmds.patch
```diff
@@ -0,0 +1,46 @@
...
++WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
++
++install(PROGRAMS "${HEKA_S3CAT_EXE}" DESTINATION bin)
++
+ add_custom_target(sbmgr ALL
+ ${GO_EXECUTABLE} install ${LDFLAGS} github.com/mozilla-services/heka/cmd/heka-sbmgr
+ DEPENDS hekad)
+-- 
+1.9.4 (Apple Git-50.2)
+
```
trink, line 46: I'll see what I cat do about getting rid of a bunch of these patch files.

heka/plugins/s3splitfile/all_specs_test.go
```diff
@@ -0,0 +1,30 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# The Initial Developer of the Original Code is the Mozilla Foundation.
+# Portions created by the Initial Developer are Copyright (C) 2012
+# the Initial Developer. All Rights Reserved.
+#
+# Contributor(s):
+#   Rob Miller (rmiller@mozilla.com)
+#   Mike Trinkala (trink@mozilla.com)
```
trink: Errr, copyright date and contributors

heka/cmd/heka-s3list/main.go
mreid-moz: The `cat` handling will give you a sensible error message if you get the parameter wrong. I will update this one to match.

heka/plugins/s3splitfile/s3splitfile_common.go
```diff
@@ -214,16 +221,224 @@ func LoadSchema(schemaFileName string) (schema Schema, err error) {
...
+	var size int64
+
+	done := false
+	for !done {
+		n, record, err := sRunner.GetRecordFromStream(reader)
+		size += int64(n)
+
+		if err != nil {
+			if err == io.EOF {
+				if len(record) != 0 {
```
trink: If the record was not empty why would it be overwritten with any remaining bytes?

heka/plugins/s3splitfile/s3splitfile_input.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	wg.Done()
+}
+
+func init() {
+	pipeline.RegisterPlugin("S3SplitFileInput", func() interface{} {
+		return new(S3SplitFileInput)
+	})
+}
```
trink, line 208: It would be nice to have a ReportMsg interface to get some visibility into its operation.

heka/cmd/heka-s3cat/main.go
mreid-moz, line 100: Within the pipeline context you don't need to do it yourself - it's just that these cmds don't have decoders.

heka/cmd/heka-s3cat/main.go
mreid-moz: I'll remove them all, since they're all over the place.  I'll leave the MPL blurb in though, right?

heka/patches/0003-Add-more-cmds.patch
mreid-moz, line 46: Thanks! That would be great.

heka/plugins/s3splitfile/s3splitfile_common.go
mreid-moz: I wanted to make sure that the Printf call below happened if something weird was going on with EOF and GetRemainingData().  If record _is_ empty, then we don't care what's in GetRemainingData(), since it's an incomplete message.  If it is not empty, we should use it, and _then_ check GetRemainingData (and not overwrite record at this point).  Will fix.

heka/plugins/s3splitfile/s3splitfile_input.go
mreid-moz: I just tried doing this, but I think my introduction of the "reachedEOF" field in Heka PR 1305 has changed the reusability of SplitterRunners - they now remember that they're at EOF and therefore don't give you any data after the end of the first stream they see.

mreid-moz: Here's a patch that made the SplitterRunners reusable - just call "Reset()" after each stream. https://gist.github.com/mreid-moz/9f9baa110050789b440a

heka/plugins/s3splitfile/s3splitfile_input.go
mreid-moz, line 208: Filed [bug 1130525](https://bugzilla.mozilla.org/show_bug.cgi?id=1130525) for this.

heka/plugins/s3splitfile/s3splitfile_input.go
mreid-moz: With [Heka PR 1317](https://github.com/mozilla-services/heka/pull/1317), we can now safely reuse SplitterRunners on different streams. I've updated the code to do just that - now we create a single SplitterRunner / Deliverer pair per `fetcher` instead of having a pair for each S3 file.

PR #4, opened 2015-02-11
Use aws.GetAuth for IAM credentials when available.

whd:
Use `GetAuth` to take advantage of environment variables and IAM credentials, if available, so that we don't need to hard-code credentials on instances anymore.

There's some gofmt spam, but salient changes are:

Default empty strings for command line tools key / secret key flags (if unspecified attempt to use IAM creds).

I added the bad region failure code from `heka-s3cat` to `heka-export` so that they maintain the same error codes. Exit 4 is now "authentication error" and 5 "bad region".

@mreid-moz r?

### DIFF
--- a/heka/cmd/heka-export/main.go
+++ b/heka/cmd/heka-export/main.go
@@ -14,18 +14,18 @@ package main
 import (
 	"flag"
 	"fmt"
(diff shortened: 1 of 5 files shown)


mreid-moz:
Looks good, r+.

PR #5, opened 2015-02-12
Build shared lib for lua-geoip

mreid-moz:
Checkout, build, and package the lua-geoip shared libraries.  There's some very weak OS detection that should make it work on Linux and OSX.  The GeoIP library is needed for [Bug 1129222](https://bugzilla.mozilla.org/show_bug.cgi?id=1129222)

### DIFF
--- a/bin/build_pipeline_heka.sh
+++ b/bin/build_pipeline_heka.sh
@@ -64,6 +64,10 @@ if [ ! -d lua-geoip ]; then
     # Fetch the lua geoip lib
     git clone https://github.com/agladysh/lua-geoip.git
     cd lua-geoip
+
+    # Use a known revision (current "master" as of 2015-02-12)
+    git checkout d9b36d7c70b7250a5c4e589d13c8b911df3c64fb
+
     # from 'make.sh'
     gcc -O2 -fPIC -I${LUA_INCLUDE_PATH} -c src/*.c -Isrc/ -Wall --pedantic -Werror --std=c99 -fms-extensions
 
--- a/bin/build_pipeline_heka.sh
+++ b/bin/build_pipeline_heka.sh
@@ -63,38 +63,36 @@ cd $BASE/build
 if [ ! -d lua-geoip ]; then
     # Fetch the lua geoip lib
     git clone https://github.com/agladysh/lua-geoip.git
-    cd lua-geoip
-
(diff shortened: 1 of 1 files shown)


bin/build_pipeline_heka.sh
```diff
@@ -52,6 +58,39 @@ fi
...
+    git clone https://github.com/agladysh/lua-geoip.git
+    cd lua-geoip
+    # from 'make.sh'
+    gcc -O2 -fPIC -I${LUA_INCLUDE_PATH} -c src/*.c -Isrc/ -Wall --pedantic -Werror --std=c99 -fms-extensions
+
+    UNAME=$(uname)
+    case $UNAME in
+    Darwin)
+        echo "Looks like OSX"
+        gcc -bundle -undefined dynamic_lookup database.o city.o -o city.so
```
trink: Just use the case to set the flags and have one set of gcc commands gcc $(SO_FLAGS) database.o city.o -o city.so ect

bin/build_pipeline_heka.sh
```diff
@@ -52,6 +58,40 @@ fi
...
+
+    gcc $SO_FLAGS database.o city.o -o city.so
+    gcc $SO_FLAGS database.o country.o -o country.so
+    gcc $SO_FLAGS database.o lua-geoip.o -o geoip.so
+    cd -
+fi
+
+cd $BASE/build/heka/build
+cp $BASE/build/lua-geoip/geoip.so heka/modules/
+mkdir -p heka/modules/geoip
```
trink: If you make this first you can build directly into the target directory and not have to copy

bin/build_pipeline_heka.sh
```diff
@@ -52,6 +58,40 @@ fi
 
 source build.sh
 
+echo 'Installing lua-geoip libs'
+cd $BASE/build
+if [ ! -d lua-geoip ]; then
+    # Fetch the lua geoip lib
+    git clone https://github.com/agladysh/lua-geoip.git
```
trink, line 65: We probably want to checkout a specific SHA instead of blindly taking whatever has been added.

bin/build_pipeline_heka.sh
mreid-moz: AIUI, the lua headers are not available until after you build heka.

PR #8, opened 2015-02-12
Address :trink's review feedback

mreid-moz:
This addresses feedback from @trink on Pull Request 7.

### DIFF
--- a/heka/sandbox/encoders/http_edge_encoder.lua
+++ b/heka/sandbox/encoders/http_edge_encoder.lua
@@ -123,7 +123,7 @@ function process_message()
     local num_components = #components
     if num_components > 0 then
         local dims = cfg["dimensions"]
-        if dims ~= nil and table.getn(dims) >= num_components then
+        if dims ~= nil and #dims >= num_components then
             for i=1,num_components do
                 msg.Fields[dims[i]] = components[i]
             end
--- a/heka/sandbox/encoders/http_edge_encoder.lua
+++ b/heka/sandbox/encoders/http_edge_encoder.lua
@@ -45,19 +45,22 @@ function get_geo_country()
     return country or UNK_GEO
 end
 
-local msg = {
-    Timestamp   = nil,
-    Type        = "http_edge_incoming",
-    Payload     = nil,
-    Fields      = {}
-}
-
 -- Load the namespace configuration externally.
 -- Note that if the config contains invalid JSON, we will discard any messages
 -- we receive with the following error:
 --    FATAL: process_message() function was not found
 local ok, ns_config = pcall(cjson.decode, read_config("namespace_config"))
 if not ok then return -1, ns_config end
+
+
+local msg = {
+    Timestamp   = nil,
+    Type        = "http_edge_incoming",
+    Payload     = nil,
+    EnvVersion  = nil,
+    Hostname    = nil,
+    Fields      = {}
+}
 
 function process_message()
     -- Reset Fields, since different namespaces may use different fields.
--- a/heka/sandbox/encoders/http_edge_encoder.lua
+++ b/heka/sandbox/encoders/http_edge_encoder.lua
@@ -4,34 +4,19 @@
 
 -- See https://wiki.mozilla.org/CloudServices/DataPipeline/HTTPEdgeServerSpecification
 
+require "lpeg"
 require "string"
 require "table"
 require 'geoip.city'
 
-local function split(str, pat)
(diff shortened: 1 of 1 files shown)


heka/sandbox/encoders/http_edge_encoder.lua
```diff
@@ -4,34 +4,19 @@
...
-    if last_end <= #str then
-        cap = str:sub(last_end)
-        table.insert(t, cap)
-    end
-    return t
-end
-
-function split_path(str)
-    return split(str,'[\\/]+')
+    sep = lpeg.P(sep_char)
```
trink: You won't want to build the grammar for every request

heka/sandbox/encoders/http_edge_encoder.lua
```diff
@@ -68,24 +53,25 @@ local ns_config = {
...
-    -- "incoming.telemetry.mozilla.org")
+    -- "incoming.telemetry.mozilla.org").
     msg.Fields.Host = read_message("Fields[Host]")
 
-    -- Path should be of the form ^/submit/namespace/id[/extra/path/components]$
+    -- Path should be of the form:
+    --     ^/submit/namespace/id[/extra/path/components]$
     local path = read_message("Fields[Path]")
-    local components = split_path(path)
+    local components = split_path(path, "/")
```
trink: split_path is now nil

heka/sandbox/encoders/http_edge_encoder.lua
```diff
@@ -68,24 +53,25 @@ local ns_config = {
...
 
     -- Skip this message: Not enough path components.
-    if components == nil or table.getn(components) < 3 then
+    if not components or #components < 3 then
```
trink, line 86: components can still never be nil from the current split function (i.e., the test is uncessary)

mreid-moz, line 86: Ah, wasn't sure if `lpeg.match(...)` could return nil.

heka/sandbox/encoders/http_edge_encoder.lua
```diff
@@ -68,24 +61,25 @@ local ns_config = {
 }
 
 function process_message()
-    -- Carry forward payload some incoming fields
```
trink, line 71: In general we initialize msg with all the fields we use:
1) to avoid additional allocation/table resizing on the first invocation of process_message
2) to document the message structure (this applies to Fields too although you have a bunch of dynamic ones here)

heka/sandbox/encoders/http_edge_encoder.lua
```diff
@@ -99,32 +93,30 @@ function process_message()
...
-    --       use a "special" field for this anyways.
-    --msg.Uuid = table.remove(components, 1)
+    -- This DocumentID is what we should use to de-duplicate submissions.
     msg.Fields.DocumentID = table.remove(components, 1)
 
-    local num_components = table.getn(components)
+    local num_components = #components
     if num_components > 0 then
         local dims = cfg["dimensions"]
         if dims ~= nil and table.getn(dims) >= num_components then
```
trink: use #dims instead of getn

heka/sandbox/encoders/http_edge_encoder.lua
mreid-moz, line 71: We could stop adding dynamic fields at this point - we could simply add a "PathComponents" array regardless of namespace config, and extract those dynamic fields downstream. I don't know if it's desirable to reference (or split) the namespace configuration in multiple places though.

heka/sandbox/encoders/http_edge_encoder.lua
```diff
@@ -137,11 +129,10 @@ function process_message()
         end
```
trink, line 133: Since you are reusing Fields you can leak variables into the next message

heka/sandbox/encoders/http_edge_encoder.lua
```diff
@@ -40,52 +34,56 @@ local UNK_GEO = "??"
...
+local ok, ns_config = pcall(cjson.decode, read_config("namespace_config"))
+if not ok then return -1, ns_config end
+
 local msg = {
     Timestamp   = nil,
     Type        = "http_edge_incoming",
     Payload     = nil,
+    EnvVersion  = nil,
+    Hostname    = nil,
     Fields      = {}
```
trink: Just set it to nil since it is a place holder

PR #9, opened 2015-02-13
Heka 0.9

trink:


heka/sandbox/decoders/decompress_payload.lua
```diff
@@ -0,0 +1,21 @@
...
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+require "string"
+local gzip = require "gzip"
+
+function process_message()
+    local payload = read_message("Payload")
+    local b1, b2 = string.byte(payload, 1, 2)
+
+    if b1 == 0x1f and b2 == 0x8b then
```
mreid-moz: - comment saying these are the gzip magic header bytes

heka/sandbox/outputs/redshift.lua
```diff
@@ -0,0 +1,155 @@
...
+assert (con:execute[[
+        CREATE TABLE IF NOT EXISTS logs(
+        Timestamp bigint,
+        Severity integer,
+        Pid integer,
+        Hostname varchar(126),
+        Type varchar(126),
+        Logger varchar(126),
+        EnvVersion varchar(126),
+        Payload varchar(126)
```
mreid-moz: Postgres will fold these column names to lowercase - if you want them in mixed-case, they should be enclosed in double-quotes.  If you do not want them in mixed-case, you should use the lowercase names in the CREATE TABLE statement.  And as a style nit, I like to see the SQL data types in uppercase (`VARCHAR` vs. `varchar`).

PR #10, opened 2015-02-17
Add example configuration and instructions.

mreid-moz:


whd:
It would be nice to symlink `build/heka/build/heka/bin` or its contents to the top-level bin directory, but I don't think there's a particularly good way to do this.

mreid-moz:
Yeah, `build/heka/build/heka` looks a bit ridiculous... we could add an install script that puts the binaries and shared libs into a target directory.

gregglind:
@mreid-moz looking this over!

PR #13, opened 2015-02-20
Change the HTTP Edge "encoder" to a "decoder".

mreid-moz:
This makes far more sense logically, but it's a bit less convenient
since you have to manually copy the incoming message in the decoder.

This enables outputs to use message matchers to properly route data.


trink:
Do you want the PR that would avoid the manual copy?

mreid-moz:
Nah, let's get this in first, and clean it up when the improved message API PR lands.

heka/sandbox/decoders/http_edge_decoder.lua
```diff
@@ -62,20 +76,43 @@ local msg = {
...
+        -- HttpListenInput, not the user-supplied one if we have to choose).
+        if not landfill_msg.Fields[name] then
+            landfill_msg.Fields[name] = value
+        end
+    end
+    inject_message(landfill_msg)
+
     -- Reset Fields, since different namespaces may use different fields.
-    msg.Fields = {}
+    main_msg.Fields = {}
```
trink, line 101: Why not just modify the landfill message here?

mreid-moz, line 101: It seemed clearer to keep the messages separate.

trink, line 101: Then I would at least move the data from the landfill message to the new message instead of hitting the CGO interface again.

PR #19, opened 2015-02-23
Gracefully handle large messages.

mreid-moz:
Now that max_message_size is a configuration parameter, there is a
risk that the sandbox decoder can attempt to output messages that
are too large, resulting in an error like:

FATAL: process_message() heka/sandbox/decoders/http_edge_decoder.lua:98: inject_message() output_limit exceeded

Which takes down Heka. We now wrap `inject_message` with `pcall`
to ensure we trap errors.


trink:
The patch file has not been removed and the build script has not been updated

heka/sandbox/decoders/http_edge_decoder.lua
```diff
@@ -173,6 +174,19 @@ function process_message()
     main_msg.Fields.geoCountry = get_geo_country(xff, remote_addr)
 
     -- Send new message along.
-    inject_message(main_msg)
-    return 0
+    local main_status, main_err = pcall(inject_message, main_msg)
+    local error_message = ""
+    local return_code = 0
+    if not landfill_status then
```
trink: Why is this not tested right after the injection?

mreid-moz: The idea was to try both "inject" operations to minimize any data loss, so the error reporting is delayed until the end of processing.

heka/sandbox/decoders/http_edge_decoder.lua
```diff
@@ -173,6 +174,19 @@ function process_message()
...
+        error_message = string.format("Error injecting landfill message: [%s] ", landfill_err)
+        return_code = -1
+    end
+    if not main_status then
+        error_message = error_message .. string.format("Error injecting main message: [%s]", main_err)
+        return_code = -1
+    end
+
+    if return_code == 0 then return return_code end
```
trink: Just initialize error_message to nil and you can remove this line

heka/sandbox/decoders/http_edge_decoder.lua
trink: Yeah, that is not the way it works. The failure will discard all messages so even in main was successful it will never see the light of day.

mreid-moz:
The patch/build change is in PR #18

heka/sandbox/decoders/http_edge_decoder.lua
mreid-moz: D'oh. I'll change the code to signal failure right away then.

PR #20, opened 2015-02-23
Bugzilla 1135252 Summary view of the telemetry decoder activity

trink:


heka/sandbox/filters/telemetry_decoder_view.lua
```diff
@@ -0,0 +1,106 @@
...
+--[[
+Creates a summary view of the TelemetryDecoder Statistics.
+--]]
+
+require "bloom_filter"
+require "circular_buffer"
+require "cjson"
+require "string"
+
+bf                  = bloom_filter.new(3*1e6, 0.01) -- todo up the size when not under the sandbox manager
```
mreid-moz, line 17: I don't know what this comment means...

heka/sandbox/filters/telemetry_decoder_view.lua
```diff
@@ -0,0 +1,106 @@
...
+        if not t then
+            return -1, "No Decoders found"
+        end
+
+        for i,v in ipairs(t) do
+            if not v.Name then
+                return -1, "Decoder is missing its name"
+            end
+
+            local id = string.match(v.Name, "^TelemetryKafkaInput(%d+)")
```
mreid-moz, line 67: Maybe add a `$` anchor at the end to ensure we don't accidentally match something else.

heka/sandbox/filters/telemetry_decoder_view.lua
trink, line 17: The bloom filter is only 3 million entries.  We need to up the size when it is not deployed under the sandbox manager.

mreid-moz, line 17: What does the sandbox manager have to do with the bloom filter size? Are sbmgr-managed sandboxes limited in the amount of memory they can use?

heka/sandbox/filters/telemetry_decoder_view.lua
```diff
@@ -0,0 +1,106 @@
...
+        end
+    end
+
+    return 0
+end
+
+last_cleared = nil
+
+function timer_event(ns)
+    if last_cleared and ns - last_cleared >= 1e9 * 86400 * 2 then
```
mreid-moz: extract a constant above for the duplicate window size.

heka/sandbox/filters/telemetry_decoder_view.lua
trink, line 67: Adding the rest of the decoder name "-TelemetryDecoders" doesn't buy us much since we control the names and we are already restricted to the decoder section of the JSON.

heka/sandbox/filters/telemetry_decoder_view.lua
mreid-moz, line 17: Never mind, I shoud've R'd TFM :)
https://hekad.readthedocs.org/en/dev/sandbox/index.html#sandbox-manager-filter

heka/sandbox/filters/telemetry_decoder_view.lua
mreid-moz, line 67: :+1:

heka/sandbox/filters/telemetry_decoder_view.lua
trink: I adjusted it although they probably shouldn't be the same I will call it close enough for now.  The clear is hard interval as opposed to a sliding window but I have no plans to make an array of blooms.

PR #21, opened 2015-02-24
Bugzilla 1131822 Firefox usage

trink:


whd:
The preferred abbreviation is fx: http://website-archive.mozilla.org/www.mozilla.org/firefox_releasenotes/en-US/firefox/releases/1.5.html#FAQ

mreid-moz:
Do we even need to abbreviate? `firefox_usage.lua` seems ok

heka/sandbox/filters/ff_usage.lua
```diff
@@ -0,0 +1,156 @@
...
+
+    if type(json.payload) ~= "table" then
+        return -1, "Missing payload object"
+    end
+
+    if type(json.payload.simpleMeasurements) ~= "table" then
+        return -1, "Missing payload.simpleMeasurements object"
+    end
+
+    local uptime = json.payload.simpleMeasurements.uptime
```
mreid-moz: We can't just accumulate `uptime` - IIUC, this is the time since the beginning of the session (and is not re-set for each subsession fragment).  I think the measure we should use is `subsessionLength` from the `payload.info` section of the `main` ping:
https://ci.mozilla.org/job/mozilla-central-docs/Tree_Documentation/toolkit/components/telemetry/telemetry/main-ping.html

trink: +1 my JSON trolling caught the first closest match and not the best match

heka/sandbox/filters/firefox_usage.lua
```diff
@@ -0,0 +1,156 @@
...
+    simply search for the uptime key/value anywhere in the string.  Parse will
+    JSON decoded the entire message to extract this one value.
+
+*Example Heka Configuration*
+
+.. code-block:: ini
+
+    [FirefoxUsage]
+    type = "SandboxFilter"
+    filename = "lua_filters/ff_usage.lua"
```
mreid-moz: update example filename

mreid-moz:
The different modes are just to test performance, right?  If the plan is to try them both and keep one, it's fine as-is. If we intend to keep them both around, we should factor out just the "get session duration from payload" piece, and use shared logic to deal with the timestamp and update the cbuf.

heka/sandbox/filters/firefox_usage.lua
```diff
@@ -0,0 +1,156 @@
...
+
+    if type(json.payload.info) ~= "table" then
+        return -1, "Missing payload.info object"
+    end
+
+    local uptime = json.payload.info.subsessionLength
+    if type(uptime) ~= "number" then
+        return -1, "Missing payload.info.subsessionLength"
+    end
+    uptime = uptime / 60 -- convert to hours
```
mreid-moz: From the main ping spec in [Bug 1120982](https://bugzilla.mozilla.org/show_bug.cgi?id=1120982):
`subsessionLength: <number>, // the subsession length in seconds`
So you should divide by 3600 rather than 60.

heka/sandbox/filters/firefox_usage.lua
```diff
@@ -0,0 +1,156 @@
...
+    return 0
+end
+
+local function match()
+    local json = read_message("Payload")
+    local uptime = string.match(json, '"subsessionLength":%s*(%d+%.?%d*)')
+    if not uptime then
+        return -1, "Missing uptime"
+    end
+    uptime = tonumber(uptime) / 60 -- convert to hours
```
mreid-moz: s/60/3600/

heka/sandbox/filters/firefox_usage.lua
trink: Yeah, one of the modes will go away, most likely 'parse' but we need to see the numbers first.

Err, different units on the session time versus sub session time should be a bug in the schema.

heka/sandbox/filters/firefox_usage.lua
```diff
@@ -0,0 +1,156 @@
...
+
+    if type(json.payload.info) ~= "table" then
+        return -1, "Missing payload.info object"
+    end
+
+    local uptime = json.payload.info.subsessionLength
+    if type(uptime) ~= "number" then
+        return -1, "Missing payload.info.subsessionLength"
+    end
+    uptime = uptime / 3600-- convert to hours
```
mreid-moz: nit: insert space before comment

PR #23, opened 2015-02-25
Carry forward more fields from incoming message

mreid-moz:


trink:
if we use the raw message instead (in this case raw should actually be correct).  We can avoid a number of CGO interactions and having to manually move the fields. This is a new feature so if you want to leave it as-is for now +1

``` lua
local raw = read_message("raw")
local msg = decode_message(raw)
```

mreid-moz:
Let's leave it for now and look at moving all the message-reading in the pipeline decoders to reading the raw message where possible as a separate issue.

PR #25, opened 2015-02-26
Small fixes

whd:
Looks like `Fields[DocumentID]` is actually `Fields[docId]`.

Also, make the telemetry stats bloom filter size a config option.

### DIFF
--- a/heka/sandbox/decoders/extract_telemetry_dimensions.lua
+++ b/heka/sandbox/decoders/extract_telemetry_dimensions.lua
@@ -55,7 +55,7 @@ function process_message()
     -- them later with values from the parsed JSON Payload if available.
     -- These fields should match the ones specified in the namespace_config for
     -- the "telemetry" endpoint of the HTTP Edge Server.
-    msg.Fields.docId            = read_message("Fields[DocumentID]")
+    msg.Fields.DocumentID       = read_message("Fields[DocumentID]")
     msg.Fields.docType          = read_message("Fields[reason]")
     msg.Fields.appName          = read_message("Fields[appName]")
     msg.Fields.appVersion       = read_message("Fields[appVersion]")
--- a/heka/sandbox/decoders/extract_telemetry_dimensions.lua
+++ b/heka/sandbox/decoders/extract_telemetry_dimensions.lua
@@ -55,7 +55,7 @@ function process_message()
     -- them later with values from the parsed JSON Payload if available.
     -- These fields should match the ones specified in the namespace_config for
     -- the "telemetry" endpoint of the HTTP Edge Server.
-    msg.Fields.DocumentID       = read_message("Fields[DocumentID]")
+    msg.Fields.documentId       = read_message("Fields[documentId]")
     msg.Fields.docType          = read_message("Fields[reason]")
     msg.Fields.appName          = read_message("Fields[appName]")
     msg.Fields.appVersion       = read_message("Fields[appVersion]")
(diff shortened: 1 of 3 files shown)


trink:
Hah, that naming changed yesterday,  @whd I would fix it in the decoder so our naming is consistent.

 edge -> main_msg.Fields.DocumentID = table.remove(components, 1)
 cep -> msg.Fields.docId = read_message("Fields[DocumentID]")

whd:
I've fixed the PR to update the decoder instead. Now we've got a single field that starts with a capital letter, but that probably doesn't matter.

heka/sandbox/filters/telemetry_decoder_view.lua
```diff
@@ -14,7 +14,9 @@ require "string"
 local SEC_PER_ROW   = 60
 local ROWS          = 2880
 
-bf                  = bloom_filter.new(3*1e6, 0.01) -- todo up the size when not under the sandbox manager
+local items         = read_config("bloom_items") or 3*1e6
```
trink, line 17: Altering these settings will require a preservation version to be added and since they are user configurable a little bit extra.

``` lua
_PRESERVATION_VERSION = read_config("preservation_version") or 0
_PRESERVATION_VERSION = _PRESERVATION_VERSION + 1 -- force a revision update due to internal changes
```

heka/sandbox/decoders/extract_telemetry_dimensions.lua
```diff
@@ -55,7 +55,7 @@ function process_message()
     -- them later with values from the parsed JSON Payload if available.
     -- These fields should match the ones specified in the namespace_config for
     -- the "telemetry" endpoint of the HTTP Edge Server.
-    msg.Fields.docId            = read_message("Fields[DocumentID]")
+    msg.Fields.DocumentID       = read_message("Fields[DocumentID]")
     msg.Fields.docType          = read_message("Fields[reason]")
```
trink, line 59: We should probably should make them all consistent.

heka/sandbox/filters/telemetry_decoder_view.lua
whd, line 17: I just blew away the filter state manually when I redeployed, but yes, this is the correct approach.

heka/sandbox/decoders/extract_telemetry_dimensions.lua
whd, line 59: I've replace DocumentID with documentId, which is consistent with the other fields.

heka/sandbox/decoders/extract_telemetry_dimensions.lua
```diff
@@ -55,7 +55,7 @@ function process_message()
     -- them later with values from the parsed JSON Payload if available.
     -- These fields should match the ones specified in the namespace_config for
     -- the "telemetry" endpoint of the HTTP Edge Server.
-    msg.Fields.docId            = read_message("Fields[DocumentID]")
+    msg.Fields.documentId       = read_message("Fields[documentId]")
     msg.Fields.docType          = read_message("Fields[reason]")
```
mreid-moz, line 59: We can make the other 2 mismatched field names consistent by updating the `namespace_config` to use `docType` and `appBuildId` instead of `reason` and `appBuildID` respectively.  The HTTP Edge names are set to match old telemetry naming, but we might as well update it to match v4 naming.

mreid-moz:
Landing this now, will address the other couple of fields in a separate PR

PR #26, opened 2015-02-27
Wrap the inject_message in pcall for safety.

mreid-moz:
We don't want the decoder to fail on oversized messages.


trink:
I am +1 on your size pre check to avoid getting all the way to inject just to catch it and throw away the work.

PR #28, opened 2015-02-27
Add a max-message-size parameter for heka-s3cat

mreid-moz:
This is the same change as made to heka-cat in [Heka PR 1373](https://github.com/mozilla-services/heka/pull/1373).


heka/cmd/heka-s3cat/main.go
```diff
@@ -43,6 +44,14 @@ func main() {
 		os.Exit(1)
 	}
 
+	if *flagMaxMessageSize < uint64(4294967295) {
```
trink: math.MaxUint32

PR #30, opened 2015-03-02
Sanitize the dimensions in the schema.

mreid-moz:
If you specify "saved-session", it will be sanitized as
"saved_session" when stored on S3. It should still match the same
filter when read back from S3.

Sanitize the dimensions before we use them to check whether a
value is allowed or not.

### DIFF
--- a/heka/plugins/s3splitfile/s3splitfile_common.go
+++ b/heka/plugins/s3splitfile/s3splitfile_common.go
@@ -99,7 +99,7 @@ type ListDimensionChecker struct {
 }
 
 func (ldc ListDimensionChecker) IsAllowed(v string) bool {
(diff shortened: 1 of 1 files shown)


heka/plugins/s3splitfile/s3splitfile_common.go
```diff
@@ -308,6 +308,8 @@ func FilterS3(bucket *s3.Bucket, prefix string, level int, schema Schema, kc cha
 			fmt.Printf("Error listing: %s\n", err)
 			// TODO: retry?
 			kc <- S3ListResult{s3.Key{}, err}
+			done = true
```
trink: Could just `break` here

PR #31, opened 2015-03-03
Add JSON report output to telemetry stats plugin

whd:
This adds reporting JSON to the telemetry stats plugin (which we need for the high-level dashboard). It doesn't require a preservation version update because it is simply adding another output.

### DIFF
--- a/heka/sandbox/filters/telemetry_decoder_view.lua
+++ b/heka/sandbox/filters/telemetry_decoder_view.lua
@@ -6,10 +6,12 @@
 Creates a summary view of the TelemetryDecoder Statistics.
 --]]
 
+require "os"
 require "bloom_filter"
 require "circular_buffer"
 require "cjson"
 require "string"
+require "table"
 
 local SEC_PER_ROW   = 60
 local ROWS          = 2880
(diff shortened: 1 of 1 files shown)


heka/sandbox/filters/telemetry_decoder_view.lua
```diff
@@ -24,6 +26,54 @@ local DUPLICATES    = cb:set_header(3, "Duplicates")
 id_count            = {} -- array of decoder ids and the last seen count
 id_failures         = {} -- array of decoder ids and the last seen failure count
 
+local max_entries = 24 * 180
+local sec_in_hour = 60 * 60
```
mreid-moz, line 30: Uppercase these 2 constants to match the ones above (SEC_PER_ROW, ROWS)

heka/sandbox/filters/telemetry_decoder_view.lua
```diff
@@ -106,5 +156,14 @@ function timer_event(ns)
         last_cleared = ns
     end
 
+    local e = cb:current_time()
+    local s = e - (e % (sec_in_hour * 1e9))
+    local sec = s / 1e9
+
+    set (report_json, sec, "total_pings", cb:compute("sum", TOTAL, s, e))
+    set (report_json, sec, "malformed_pings", cb:compute("sum", FAILURES, s, e))
+    set (report_json, sec, "duplicate_pings", cb:compute("sum", DUPLICATES, s, e))
```
mreid-moz, line 165: nit: remove space between `set` and `(`

mreid-moz, line 165: Consider using the same field names in JSON (`total`, `failures`, `duplicates`)

trink:
When the hour changes the data from the previous hour isn't summed again and will produce inaccurate counts.  Also since the metrics people seem to be fixated on verbose JSON output we should probably stop rolling our own circular buffer like structure every time we need JSON output and instead just make a cbuf json format.

trink:
We will extend d3 to be able to parse the cbuf output (like the js adaptor we wrote for dygraphs).

PR #32, opened 2015-03-04
Break the main ping into separate message fields

trink:


heka/sandbox/decoders/extract_telemetry_dimensions.lua
```diff
@@ -87,6 +104,13 @@ function process_message()
         msg.Fields.clientId         = parsed.clientID
     elseif parsed.version then
         -- New-style telemetry, see http://mzl.la/1zobT1S
+        if parsed.type == "main" then
+            msg.Payload = payload -- keep the gzipped payload since we are moving most of the content into fields
```
trink: Do we want to keep it? The original payload (gzipped in most cases), the full JSON, the remaining JSON after the object extraction?

mreid-moz: I would prefer to see just the remaining JSON so we don't store too much duplicate data.

heka/sandbox/decoders/extract_telemetry_dimensions.lua
```diff
@@ -12,13 +12,29 @@ local msg = {
...
 Type        = "telemetry",
 Payload     = nil,
-Fields      = { sourceName = "telemetry" }
+EnvVersion  = 1
 }
 
 local UNK_DIM = "UNKNOWN"
 local UNK_GEO = "??"
+local environment_objects = {"build", "settings", "profile", "partner", "system", "addons"}
+local main_ping_objects = {"info", "simpleMeasurements", "histograms", "keyedHistograms", "chromeHangs", "threadHangStats", "log", "fileIOReports", "lateWrites", "addonDetails", "addonHistograms", "UIMeasurements", "slowSQL", "slowSQLstartup", "childPayloads"}
```
mreid-moz: Nit: split this long list into multiple lines for readabillity, and ideally sort the object names alphabetically (unless the current order has significance).

heka/sandbox/decoders/extract_telemetry_dimensions.lua
```diff
@@ -12,13 +12,29 @@ local msg = {
...
+
+local function split_objects(root, section, objects)
+    if type(root) ~= "table" then return end
+
+    for i, name in ipairs(objects) do
+        if type(root[name]) == "table" then
+            local ok, json = pcall(cjson.encode, root[name])
+            if ok then
+                msg.Fields[string.format("%s.%s", section, name)] = json
```
mreid-moz, line 55: Should we check to make sure we're not overwriting an existing field name? Given the `section.name` scheme, that is unlikely, but it may be good to detect that case.

heka/sandbox/decoders/extract_telemetry_dimensions.lua
trink: They were just ordered as they appear.
Split out one per line and sorted.

heka/sandbox/decoders/extract_telemetry_dimensions.lua
trink, line 55: We are in full control of the message, they don't exist.  Also, even if they did I believe the correct behaviour would be to overwrite otherwise they could be hijacked (unless you want to fail the message?).

mreid-moz, line 55: Ok, as long as we're sure overwriting is the right thing to do, I'm good with this.

PR #33, opened 2015-03-04
Remove the default invalid bucket.

mreid-moz:
There's no reason to default to something that won't work.


trink:
This will still crash if the now provided bucket name is invalid.

mreid-moz:
I can live with a crash if you use a bad name (at least for now). I need to do a full pass over all the S3 code to make it more robust and safe. Can we punt until then?

trink:
r-
This doesn't buy us much as-is and should be addressed in the general clean up.

PR #38, opened 2015-03-12
Add the creation date of the telemetry ping to the fields.

vitillo:


heka/sandbox/decoders/extract_telemetry_dimensions.lua
```diff
@@ -102,6 +102,7 @@ function process_message()
         msg.Fields.appBuildId       = app.buildId
         msg.Fields.vendor           = app.vendor
         msg.Fields.clientId         = parsed.clientId
+        msg.Fields.creationDate     = parsed.creationDate
```
mreid-moz, line 174: Should we convert this value to the same form as we use for the `msg.Timestamp` field? The incoming value is an ISO 8601 date.

vitillo, line 174: I thought about it but that would create a mismatch between the payload and the fields. Ideally we should preserve the format of the fields in the payload to avoid surprises in analyses downstream.

mreid-moz, line 174: If our aim is to be able to use it for calculations, then we might as well parse it - if you want the "raw" string version you can always parse it out of the payload...  I agree with avoiding surprises downstream, but having all the timestamps in the same format is unsurprising in other ways :)

vitillo, line 174: I was more thinking on the line that hopefully in the future, when we have a schema, we will extract more fields from the json blob and whoever does the analysis will expect those fields to match the schema. We could just rename it to creationDateTimestamp and adopt the same format as msg.Timestamp though.

vitillo:
I have updated the field as discussed on IRC.

heka/sandbox/decoders/extract_telemetry_dimensions.lua
```diff
@@ -81,6 +82,15 @@ local function sample(id, sampleRange)
     return hash.crc32(id) % sampleRange
 end
 
+function parse_creation_date(date)
+   local t = dt.rfc3339:match(date)
```
mreid-moz, line 88: Does this cause an error if `date` is nil? If so, please check that first.

heka/sandbox/decoders/extract_telemetry_dimensions.lua
```diff
@@ -81,6 +82,15 @@ local function sample(id, sampleRange)
     return hash.crc32(id) % sampleRange
 end
 
+function parse_creation_date(date)
+   local t = dt.rfc3339:match(date or "")
```
mreid-moz: Don't bother to try and parse if the date is missing. Please guard with something like `if not date or type(date) ~= "string" then return nil end`

trink: This will suffice:

``` lua
if type(date) ~= "string" then return nil end
```

However, creationDate is not optional so we should failed the message if it doesn't exist or fails to parse.

heka/sandbox/decoders/extract_telemetry_dimensions.lua
```diff
@@ -81,6 +82,18 @@ local function sample(id, sampleRange)
...
 
+function parse_creation_date(date)
+   if type(date) ~= "string" then return nil end
+
+   local t = dt.rfc3339:match(date)
+   if not t then
+      return nil
+   end
+
+   ns = dt.time_to_ns(t) -- The timezone of the ping has always zero UTC offset
```
trink: ns is global... instead just return this directly

PR #44, opened 2015-03-19
Add telemetry latency job.

vitillo:


heka/sandbox/decoders/extract_telemetry_dimensions.lua
```diff
@@ -121,14 +121,14 @@ function process_message()
     msg.Fields.appUpdateChannel = read_message("Fields[appUpdateChannel]")
     msg.Fields.appBuildId       = read_message("Fields[appBuildId]")
 
+    local info = parsed.info
```
trink: `info` is not a top level object in the new telemetry format it is actually `parsed.payload.info`.  This will reject all new messages.

heka/sandbox/filters/telemetry_latency.lua
```diff
@@ -0,0 +1,224 @@
...
+   - An active profile is one who has used the browser in the last six weeks (42 days)
+
+*Example Heka Configuration*
+
+.. code-block:: ini
+
+    [TelemetryLatency]
+    type = "SandboxFilter"
+    filename = "lua_filters/telemetry_latency.lua"
+    message_matcher = "Type == 'telemetry' && Fields[docType] == 'main'"
```
trink: You should accept on version and sample_id in the matcher.  It will save a trip down to the sandbox and the reads+test can be removed from the plugin.

vitillo: Great suggestion!

heka/sandbox/filters/telemetry_latency.lua
```diff
@@ -0,0 +1,224 @@
...
+   if version == PING_VERSION and sample_id == 0 then
+      local ts = read_message("Timestamp")
+      local client_id = read_message("Fields[clientId]")
+      local creation_ts = read_message("Fields[creationTimestamp]") -- exists only in new "unified" pings
+      local channel = read_message("Fields[appUpdateChannel]") or "UNKNOWN"
+      local reason = read_message("Fields[reason]")
+      local ch_reason = channel .. ":" .. reason
+
+      process_client_metric(seen_by_channel, ch_reason, client_id, ts)
+      process_client_metric(creation_delay_by_channel, ch_reason, client_id, ts - creation_ts)
```
trink: The client metrics tables could be consolidated (it would save space and lookups).  The same for the subsession metrics.  If the size or speed is a problem.

heka/sandbox/filters/telemetry_latency.lua
```diff
@@ -0,0 +1,224 @@
...
+   local sample_id = read_message("Fields[sampleId]")
+   local version = read_message("Fields[sourceVersion]")
+
+   if version == PING_VERSION and sample_id == 0 then
+      local ts = read_message("Timestamp")
+      local client_id = read_message("Fields[clientId]")
+      local creation_ts = read_message("Fields[creationTimestamp]") -- exists only in new "unified" pings
+      local channel = read_message("Fields[appUpdateChannel]") or "UNKNOWN"
+      local reason = read_message("Fields[reason]")
+      local ch_reason = channel .. ":" .. reason
```
trink: This is unbounded, we probably want to limit the dimensions so the plugin cannot be terminated by unexpected input.

vitillo: We can address this by matching only the nightly channel in the configuration of the filter.

trink:
I am fine with getting it merged but if it has specific configuration requirements (memory size and channel filter) they should be captured/noted in the documentation.  For now just set it to something like 200MiB so people are aware it is a potentially large one.

PR #48, opened 2015-04-07
Move the message restructuring to the end of the block

trink:
- also fail faster by moving the schema tests to the top of the block

### DIFF
--- a/heka/sandbox/decoders/extract_telemetry_dimensions.lua
+++ b/heka/sandbox/decoders/extract_telemetry_dimensions.lua
@@ -143,20 +143,31 @@ function process_message()
         msg.Fields.clientId         = parsed.clientID
     elseif parsed.version then
         -- New-style telemetry, see http://mzl.la/1zobT1S
-        if parsed.type == "main" then
-            split_objects(parsed.environment, "environment", environment_objects)
-            split_objects(parsed.payload, "payload", main_ping_objects)
-            local ok, json = pcall(cjson.encode, parsed) -- re-encode the remaining data
-            if not ok then return -1, json end
-            msg.Payload = json
-        else
-            msg.Payload = json
+
+        -- pull out/verify the data/schema before any restructuring
+        local app = parsed.application
+        if type(app) ~= "table" then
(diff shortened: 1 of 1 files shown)


heka/sandbox/decoders/extract_telemetry_dimensions.lua
```diff
@@ -143,21 +143,32 @@ function process_message()
...
-            msg.Payload = json
+
+        -- pull out/verify the data/schema before any restructuring
+        local app = parsed.application
+        if type(app) ~= "table" then
+            return -1, "missing 'application' object"
+        end
+
+        if type(parsed.payload.info) ~= "table" then
+           return -1, "missing info object"
```
mreid-moz: We should be consistent with single-quoting object names (or not) - the messages for missing `application`, `info`, and `creationDate` should look similar.

mreid-moz: I'm not sure we care enough about the `reason` field to make it a decode error if it is missing.

heka/sandbox/decoders/extract_telemetry_dimensions.lua
```diff
@@ -143,21 +143,32 @@ function process_message()
...
-            msg.Payload = json
+
+        -- pull out/verify the data/schema before any restructuring
+        local app = parsed.application
+        if type(app) ~= "table" then
+            return -1, "missing application object"
+        end
+
+        if type(parsed.payload.info) ~= "table" then
+           return -1, "missing info object"
```
mreid-moz: Messages look good now, but I still think we should change this case not to be an error (like `os` below, just use it if it's available).

trink: If it is missing the info OBJECT it should be an error (granted we are only trying to extract reason from it)

mreid-moz: Right, I don't necessarily want to force all future ping types to include an `info` object if they don't have a `reason` field.

PR #56, opened 2015-04-09
The 'other' channel was being skipped

trink:
- added aurora in, basically for the new count only


heka/sandbox/filters/firefox_channel_switching.lua
```diff
@@ -49,8 +50,8 @@ channels = {
     {name = "beta"           , cb = create_cbuf(), cf = cuckoo_filter.new(10e6)},
     {name = "nightly"        , cb = create_cbuf(), cf = cuckoo_filter.new(1e6)},
 
-    -- skipping aurora since it uses a different profile we cannot track the switches
-    -- {name = "aurora", cb = create_cbuf(), cf = cuckoo_filter.new(1e6)},
+    -- aurora uses a different profile so we should not be able to track the switches
```
mreid-moz: "we do not expect to see any switches"?

PR #55, opened 2015-04-09
Add an errors only stream for the telemetry data

trink:

PR #57, opened 2015-04-10
Bugzilla 1149666

trink:


mreid-moz:
Can you rebase this off `master`?

trink:
Why only the last diff is different?

mreid-moz:
To get rid of the duplicated commits from #55.

heka/sandbox/encoders/combine_telemetry_objects.lua
```diff
@@ -0,0 +1,34 @@
...
+            local ok, object = pcall(cjson.decode, msg.Fields[i].value[1])
+            if ok then
+                json[section][name] = object
+            end
+        end
+    end
+    local ok, payload = pcall(cjson.encode, json)
+    if not ok then return -1, payload end
+
+    inject_payload("txt", "output", msg.clientId, "\t", payload, "\n")
```
mreid-moz, line 32: The previous data export included the leftover Fields in the output - there are a few things that are not inside the payload that could be useful, though I'm not sure whether they are neccessary or not. `geoCountry` and `Timestamp` are likely to be of interest.

trink:
It is only a github display issue.

heka/sandbox/encoders/combine_telemetry_objects.lua
trink, line 32: The goal is to pass back the original structure, correct?

mreid-moz:
I get an error of "error: Failed serializing: bad argument #0 to '?' (must have one string argument)" when trying to use this with the S3 input - is that due to the "raw" message not being properly assigned in the input code?

heka/sandbox/encoders/combine_telemetry_objects.lua
mreid-moz, line 32: I would like to provide everything in the heka message, with the payload combined back into its original structure.

trink:
Yeah, this PR pulls in the Heka with the fix (i.e. you need to build the pipeline with this tag for this encoder to work)

PR #60, opened 2015-04-14
Get specific records by offset

mreid-moz:
This PR includes two main changes:
1. Parallelize heka-s3cat to improve throughput.
2. Add a `get-clients` cmd which can fetch specific records within S3 files using a local index.  This is intended to be used to fetch records for a particular `clientId` for telemetry data.

It also adds an "offsets" output format to `heka-s3cat` for generating the index expected by `get-clients`.


mreid-moz:
I need to rebase this on the "AdRoll" change from #61.  Will submit a new PR when that's ready.

PR #61, opened 2015-04-15
Update goamz library from crowdmob to AdRoll

mreid-moz:


bin/build_pipeline_heka.sh
```diff
@@ -123,6 +125,10 @@ case $UNAME in
 Darwin)
     # Don't bother trying to build a package on OSX
     make
+    echo "If you see an error like:"
```
trink, line 131: Might just want to export it for them

mreid-moz, line 131: Wouldn't help unless you source the script instead of calling it in a sub-shell

trink, line 131: Yeah but you know it will fail every time on a new build. Also, the same issue should exist on Linux, just LD_LIBRARY_PATH instead.

PR #64, opened 2015-04-20
Parallelize heka-s3cat

mreid-moz:
Fetch multiple files at the same time to improve throughput, and add an "offsets" mode for indexing the records within the S3 data files by client id.


heka/cmd/heka-s3cat/main.go
```diff
@@ -53,6 +54,16 @@ func main() {
 		os.Exit(8)
 	}
 
+	workers := 1
+	if *flagWorkers == 0 {
+		fmt.Printf("Cannot run with zero workers. Using 1.\n")
+	} else if *flagWorkers < math.MaxUint32 {
```
trink: That is a lot of worker may want to cap it in the hundreds instead of the billions

PR #69, opened 2015-05-13
Add an example monitor: DNT Status

mreid-moz:
Checks the status of the Do Not Track HTTP Header.

### DIFF
--- a/examples/monitor_dnt.lua
+++ b/examples/monitor_dnt.lua
@@ -16,8 +16,12 @@ Monitor DNT header status
     ticker_interval = 10
     preserve_data = true
         [DNTUsage.config]
+        # Increment this if the format changes in a
+        # backwards-incompatible way
         preservation_version = 1
+        # Number of entries to keep in the circular buffer
         rows = 1440
+        # Length of each bucket in the circular buffer
         sec_per_row = 300
 
 --]]
@@ -25,6 +29,7 @@ _PRESERVATION_VERSION = read_config("preservation_version") or 0
 
 require "circular_buffer"
 
+-- Default to 2880 minute-long intervals
 local rows = read_config("rows") or 2880
 local sec_per_row = read_config("sec_per_row") or 60
 
@@ -54,5 +59,9 @@ function process_message ()
 end
 
 function timer_event(ns)
-    inject_payload("cbuf", "DNT Status", c)
+    -- Inject the entire circular buffer
+    inject_payload("cbuf", "DNT Status", c:format("cbuf"))
+
+    -- Inject the cbuf delta (changes since last timer event)
+    inject_payload("cbufd", "DNT Status", c:format("cbufd"))
 end
--- a/examples/monitor_dnt.lua
+++ b/examples/monitor_dnt.lua
@@ -9,21 +9,29 @@ Monitor DNT header status
 
 .. code-block:: ini
 
-    [DNT Usage]
+    [DNTUsage]
     type = "SandboxFilter"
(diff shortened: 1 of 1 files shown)


examples/monitor_dnt.lua
```diff
@@ -0,0 +1,50 @@
...
+local rows = read_config("rows") or 2880
+local sec_per_row = read_config("sec_per_row") or 60
+local ON  = 1
+local OFF = 2
+local UNK = 3
+
+local c = circular_buffer.new(rows, 3, sec_per_row, true)
+c:set_header(ON, "DNT On")
+c:set_header(OFF, "DNT Off")
+c:set_header(UNK, "DNT Unknown")
```
trink:
``` lua
-- remove constants

c:set_header(1, "DNT Off")
c:set_header(2, "DNT On")
c:set_header(3, "DNT Unknown")
```

...

``` lua
local item = tonumber(read_message("Fields[DNT]")) or 2

-- remove if
-- range check or pcall
c:add(ts, item+1, 1)
```

mreid-moz: Some of the motivation for this filter is to serve as an example - I think the general pattern of variables for cbuf column numbers + explicitly setting their names makes a good model.

trink: Maybe too compact for an example... 
You can just consolidate lines 24-26 with 29-31
i.e., 'local ON = c:set_header(1, "DNT On")

and you can get rid of the 'or "UNKNOWN"'

examples/monitor_dnt.lua
```diff
@@ -0,0 +1,50 @@
...
+    preserve_data = true
+--]]
+
+require "circular_buffer"
+
+local rows = read_config("rows") or 2880
+local sec_per_row = read_config("sec_per_row") or 60
+
+-- Create a circular buffer with three columns
+local c = circular_buffer.new(rows, 3, sec_per_row, true)
```
trink: This would have to be global to be preserved

examples/monitor_dnt.lua
```diff
@@ -0,0 +1,50 @@
...
+.. code-block:: ini
+
+    [DNT Usage]
+    type = "SandboxFilter"
+    filename = "examples/monitor_dnt.lua"
+    message_matcher = "Type == 'telemetry'"
+    ticker_interval = 10
+    preserve_data = true
+--]]
+
```
trink, line 29:
``` lua
_PRESERVATION_VERSION = read_config("preservation_version") or 0
```

examples/monitor_dnt.lua
```diff
@@ -0,0 +1,58 @@
...
+        c:add(ts, OFF, 1)
+    else
+        c:add(ts, UNK, 1)
+    end
+
+    return 0
+end
+
+function timer_event(ns)
+    inject_payload("cbuf", "DNT Status", c)
```
trink: What about injecting the cbuf delta?

examples/monitor_dnt.lua
```diff
@@ -0,0 +1,58 @@
...
+    [DNTUsage]
+    type = "SandboxFilter"
+    filename = "examples/monitor_dnt.lua"
+    message_matcher = "Type == 'telemetry'"
+    ticker_interval = 10
+    preserve_data = true
+        [DNTUsage.config]
+        preservation_version = 1
+        rows = 1440
+        sec_per_row = 300
```
trink, line 25: Since this is an example we probably want docs explaining what the options do.

PR #70, opened 2015-05-15
Support filtering by object basename prefix in S3SplitFileInput

whd:
```
[12:12:19] <whd> mreid: we didn't add "filter by date" type functionality to s3 input, did we?
[12:12:56] <mreid> whd: just by dimension
[12:13:24] <whd> so not by file name e.g. for landfill
[12:13:56] <mreid> whd: no, but PRs accepted :)
```

This is my current workaround for the SIGPIPE issue we're seeing, and is also useful for the single-day cutover backfill.

### DIFF
--- a/heka/plugins/s3splitfile/s3splitfile_input.go
+++ b/heka/plugins/s3splitfile/s3splitfile_input.go
@@ -13,6 +13,7 @@ import (
 	"github.com/mozilla-services/heka/message"
 	"github.com/mozilla-services/heka/pipeline"
 	"io"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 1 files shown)


heka/plugins/s3splitfile/s3splitfile_input.go
```diff
@@ -39,27 +40,29 @@ type S3SplitFileInputConfig struct {
...
-	S3BucketPrefix string `toml:"s3_bucket_prefix"`
-	S3Retries      uint32 `toml:"s3_retries"`
-	S3WorkerCount  uint32 `toml:"s3_worker_count"`
+	SchemaFile             string `toml:"schema_file"`
+	AWSKey                 string `toml:"aws_key"`
+	AWSSecretKey           string `toml:"aws_secret_key"`
+	AWSRegion              string `toml:"aws_region"`
+	S3Bucket               string `toml:"s3_bucket"`
+	S3BucketPrefix         string `toml:"s3_bucket_prefix"`
+	S3ObjectBasenamePrefix string `toml:"s3_object_basename_prefix"`
```
mreid-moz: Should we add a basename "pattern" instead of a simple "prefix"?  It should be about the same complexity, but much more powerful.

mreid-moz:
Other than using a pattern instead of a prefix, this looks good to me.

PR #74, opened 2015-05-28
Telemetry elasticsearch

whd:
The es_fields encoder here was meant to be generic for mozilla-services/heka#873 but I haven't figured out how we want to handle timestamps so there's some special case logic there. Latency (a derived field) can be calculated either in the telemetry decoder or the elasticsearch encoder, but Size needs to happen in the decoder unless we want to do a lot of extra work.

I'm using capitalized field names for fields computed on the server, which I think was the convention we chose.

We might want Latency in s or ms instead of ns.

### DIFF
--- a/heka/sandbox/encoders/es_fields.lua
+++ b/heka/sandbox/encoders/es_fields.lua
@@ -1,0 +1,140 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+--[[
+Elasticsearch encoder to index only specific fields.
+
+Config:
+
+- index (string, optional, default "heka-%{%Y.%m.%d}")
+    String to use as the `_index` key's value in the generated JSON. Supports
+    field interpolation as described below.
+- type_name (string, optional, default "message")
+    String to use as the `_type` key's value in the generated JSON. Supports
+    field interpolation as described below.
+- id (string, optional)
+    String to use as the `_id` key's value in the generated JSON. Supports
+    field interpolation as described below.
+- es_index_from_timestamp (boolean, optional)
+    If true, then any time interpolation (often used to generate the
+    ElasticSeach index) will use the timestamp from the processed message
+    rather than the system time.
+- fields (string, required)
+    JSON array of fields to index.
+
+Field interpolation:
+
+    Data from the current message can be interpolated into any of the string
+    arguments listed above. A `%{}` enclosed field name will be replaced by
+    the field value from the current message. Supported default field names
+    are "Type", "Hostname", "Pid", "UUID", "Logger", "EnvVersion", and
+    "Severity". Any other values will be checked against the defined dynamic
+    message fields. If no field matches, then a `C strftime
+    <http://man7.org/linux/man-pages/man3/strftime.3.html>`_ (on non-Windows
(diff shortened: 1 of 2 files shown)


heka/sandbox/encoders/es_fields.lua
```diff
@@ -0,0 +1,147 @@
...
+    Severity = "Severity",
+    Timestamp = "Timestamp",
+    Logger = "Logger"
+}
+
+local static_fields = {}
+local dynamic_fields = {}
+
+local function key(str)
+    return str:match("Fields%[(.+)%]") or str
```
trink: This should fail as a configuration error - invalid field name (it will be a silent failure and a wasted read_message that returns nil)

heka/sandbox/encoders/es_fields.lua
```diff
@@ -0,0 +1,147 @@
...
+    end
+
+    local idx_json = elasticsearch.bulkapi_index_json(index, type_name, id, ns)
+
+    local tbl = {}
+    for i, field in ipairs(static_fields) do
+        if field == "Timestamp" and ts_from_message then
+            tbl[field] = os.date("!%Y-%m-%dT%XZ", ns / 1e9)
+        else
+            tbl[field] = read_message(field)
```
trink, line 117: This will set the Timestamp to the nanosecond value of the message and not the system time as documented.

heka/sandbox/encoders/es_fields.lua
```diff
@@ -0,0 +1,147 @@
...
+        while v do
+            if z == 0 then
+                tbl[field] = v
+            elseif z == 1 then
+                tbl[field] = {tbl[field], v}
+            elseif z > 1 then
+                tbl[field][#tbl[field]+1] = v
+            end
+            z = z + 1
+            v = read_message(field, nil, z)
```
trink: Should be f, not field

heka/sandbox/encoders/es_fields.lua
```diff
@@ -0,0 +1,147 @@
...
+        end
+    end
+
+    for i, field in ipairs(dynamic_fields) do
+        local f = string.format("Fields[%s]", field)
+        local z = 0
+        local v = read_message(f, nil, z)
+        while v do
+            if z == 0 then
+                tbl[field] = v
```
trink, line 127: Doesn't handle collisions between static and dynamic fields.  The dynamic field would clobber the header field.

heka/sandbox/encoders/es_fields.lua
```diff
@@ -0,0 +1,147 @@
...
+        static_fields[#static_fields+1] = field
+    else
+        dynamic_fields[#dynamic_fields+1] = key(field)
+    end
+end
+
+function process_message()
+    local ns
+    if ts_from_message then
+        ns = read_message("Timestamp")
```
trink, line 107: Pulling down the entire message may be faster (depending on how many fields are actually consumed).  You would also have access to the size of the entire message at that point. However, as-is is probably optimal for telemetry data.

heka/sandbox/encoders/es_fields.lua
whd, line 127: AFAIK we don't handle this in the standard elasticsearch encoder either, so I don't think we should worry about it here.

heka/sandbox/encoders/es_fields.lua
whd, line 107: I imagine the size would be slightly different from the size of the JSON, but probably comparable. We could have a config option to pull down the entire message but I don't think it matters in our case.

PR #75, opened 2015-05-29
Bugzilla #1168412 Duplicate Telemetry Submission Report

trink:


heka/sandbox/filters/firefox_duplicates.lua
```diff
@@ -0,0 +1,74 @@
...
+        items = 100000000
+--]]
+
+require "bloom_filter"
+require "circular_buffer"
+local fx = require "fx"
+
+local items = read_config("items") or 1000000
+local probability = read_config("probability") or 0.01
+bf = bloom_filter.new(items, probability)
```
mreid-moz, line 33: What will happen if you try to create a bloom filter that requires more memory than is available?

heka/sandbox/filters/firefox_duplicates.lua
```diff
@@ -0,0 +1,74 @@
...
+        [FirefoxDuplicates.config]
+        items = 100000000
+--]]
+
+require "bloom_filter"
+require "circular_buffer"
+local fx = require "fx"
+
+local items = read_config("items") or 1000000
+local probability = read_config("probability") or 0.01
```
mreid-moz, line 32: Do we need to check that 0 < probability < 1?

heka/sandbox/filters/firefox_duplicates.lua
trink, line 33: Same as any other application, it will swap like crazy and end badly.

heka/sandbox/filters/firefox_duplicates.lua
trink, line 32: Already done https://github.com/mozilla-services/lua_bloom_filter/blob/master/lua_bloom_filter.c#L43

heka/sandbox/filters/firefox_duplicates.lua
```diff
@@ -0,0 +1,74 @@
...
+end
+
+cids = {}
+
+function process_message()
+    local did = read_message("Fields[documentId]")
+    if type(did) == "string" then
+        if not bf:add(did) then
+            local ts  = read_message("Timestamp")
+            local channel = read_message("Fields[channel]")
```
mreid-moz, line 48: Should use the normalized channel here

trink, line 48: It is normalized on L49.  This runs against the curated executive summary stream so if someone injects garbage it will just default to Other.

mreid-moz, line 48: I mean we should run `fx.normalize_channel` on it, but if this plugin always runs against the executive report stream (as specified in the example config on line 15) then we don't need to.

PR #76, opened 2015-06-04
Track latency in seconds

whd:
...as nanoseconds is a bit much.


whd:
As this is derived data for kibana (where we can't modify the presentation layer) I can move this logic to es_fields.lua instead. If people want this number in analysis they can perform the computation there to the precision they desire.

whd:
Updated to calculate latency in es_fields, but as per usual failing travis.

PR #77, opened 2015-06-05
Generate an alert if the telemetry data drops to zero

trink:


heka/sandbox/filters/telemetry_decoder_view.lua
```diff
@@ -106,5 +119,13 @@ function timer_event(ns)
         last_cleared = ns
     end
 
+    if not cb:get(ns, 1) then
+        cb:add(ns, 1, 0/0) -- always advance the buffer/graph
```
mreid-moz: what is the `0/0` thing?

trink: This creates a new entry in the circular buffer with the default value of NaN.

mreid-moz: Can you update the comment to `always advance the buffer/graph, defaulting to NaN`?

PR #78, opened 2015-06-05
Add the ProcessFileFailures monitor for S3Outputs

trink:


heka/sandbox/filters/telemetry_s3output_monitors.lua
```diff
@@ -0,0 +1,68 @@
...
+    if type(json.outputs) ~= "table" then return -1, "missing outputs array" end
+
+    for i,v in ipairs(json.outputs) do
+        if type(v) ~= "table" then return -1, "invalid output object" end
+        if type(v.ProcessFileFailures) == "table" then -- confirm this plugin has instrumentation
+            if not v.Name then return -1, "missing plugin Name" end
+            local p = find_plugin(v.Name)
+            local n = v.ProcessFileFailures.value
+            if type(n) == "number" and n > p.last_value then
+                p.msg = string.format("%s ProcessFileFailures has increased to %d", v.Name, n)
```
mreid-moz, line 49: Can we add "increased from %d to %d" to show the delta from `last_value` as well?

trink, line 49: It wasn't included because it won't be the delta if it increments multiple times in a minute or if we suppress alerts.

mreid-moz, line 49: Ok, that makes sense.

Issue #81, opened 2015-06-12
S3SplitFileInput fetchers are hanging


trink:
In this case 4 of the 16 were hung down in the split_runner.  The only logged errors were about 150 connection resets.

2015/06/12 03:58:14 Input 'TelemetryInput' error: Error reading telemetry-2/20150611/telemetry/4/main/Firefox/beta/39.0/20150609130336/20150611191821.676_ip-172-31-14-40: read tcp 54.231.160.88:443: connection reset by peer

Looks like they are blocked on read down in the split runner see:
goroutine 178474 [IO wait, 821 minutes]:
goroutine 324166 [IO wait, 625 minutes]:
goroutine 324166 [IO wait, 625 minutes]:
goroutine 172972 [IO wait, 860 minutes]:

```
^\SIGQUIT: quit
PC=0x4470d1

goroutine 0 [idle]:
runtime.futex(0x1028ed8, 0x0, 0x0, 0x0, 0x0, 0x1028320, 0x1, 0x414aff, 0x414d1e, 0x1028ed8, ...)
    /usr/local/go/src/runtime/sys_linux_amd64.s:277 +0x21
runtime.futexsleep(0x1028ed8, 0x0, 0xffffffffffffffff)
    /usr/local/go/src/runtime/os_linux.c:49 +0x47
runtime.notesleep(0x1028ed8)
    /usr/local/go/src/runtime/lock_futex.go:145 +0xae
stopm()
    /usr/local/go/src/runtime/proc.c:1178 +0x119
startlockedm(0xc2080010e0)
    /usr/local/go/src/runtime/proc.c:1329 +0xa8
schedule()
    /usr/local/go/src/runtime/proc.c:1582 +0x17a
runtime.park_m(0xc2080a2fc0)
    /usr/local/go/src/runtime/proc.c:1654 +0x113
runtime.mcall(0x43d6b4)
    /usr/local/go/src/runtime/asm_amd64.s:186 +0x5a

goroutine 1 [semacquire]:
sync.(*WaitGroup).Wait(0xc20808c7a0)
    /usr/local/go/src/sync/waitgroup.go:132 +0x169
github.com/mozilla-services/heka/pipeline.Run(0xc20808c600)
    /work/git/data-pipeline/build/heka/build/heka/src/github.com/mozilla-services/heka/pipeline/pipeline_runner.go:327 +0x1fa2
main.main()
[...]

PR #82, opened 2015-06-12
Introduce config params for S3 timeouts.

mreid-moz:
This addresses #81

### DIFF
--- a/heka/plugins/s3splitfile/s3splitfile_input.go
+++ b/heka/plugins/s3splitfile/s3splitfile_input.go
@@ -50,6 +50,8 @@ type S3SplitFileInputConfig struct {
 	S3BucketPrefix     string `toml:"s3_bucket_prefix"`
 	S3ObjectMatchRegex string `toml:"s3_object_match_regex"`
 	S3Retries          uint32 `toml:"s3_retries"`
+	S3ConnectTimeout   uint32 `toml:"s3_connect_timeout"`
+	S3ReadTimeout      uint32 `toml:"s3_read_timeout"`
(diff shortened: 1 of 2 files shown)


heka/plugins/s3splitfile/s3splitfile_input.go
```diff
@@ -64,6 +66,8 @@ func (input *S3SplitFileInput) ConfigStruct() interface{} {
 		S3BucketPrefix:     "",
 		S3ObjectMatchRegex: "",
 		S3Retries:          5,
+		S3ConnectTimeout:   120,
+		S3ReadTimeout:      120,
```
trink: 2 minutes seem excessive since it will block shutdown.  @whd opinions?

heka/plugins/s3splitfile/s3splitfile_output.go
```diff
@@ -114,19 +116,21 @@ const (
...
+		MaxFileSize:      524288000,
+		MaxFileAge:       3600000,
+		MaxOpenFiles:     1000,
+		AWSKey:           "",
+		AWSSecretKey:     "",
+		AWSRegion:        "us-west-2",
+		S3Bucket:         "",
+		S3BucketPrefix:   "",
+		S3Retries:        5,
+		S3ConnectTimeout: 300,
```
trink: Why are the output defaults different?

mreid-moz: Failures in S3Output are far more important than S3Input. We want to wait as long as possible before declaring failure. There is actual retry logic in the output (where that is not implemented in the input yet), so this should mostly be mitigated by retries, but tl;dr we want to try harder on Output.

PR #83, opened 2015-06-12
Retry s3input read

mreid-moz:
If there is an error reading from S3, retry from the last known-good offset.

Note: I have not yet fully tested that this produces the exact same results as without the retry logic.


mreid-moz:
I just ran a test on 30,982 files (2,029,974.94MB) and got exactly the expected results (with a few retries during the run).  I am happy with that, feel free to test further if you like.

PR #91, opened 2015-07-06
Rename geo to country

trink:


heka/sandbox/filters/firefox_monthly_dashboard.lua
```diff
@@ -91,7 +91,7 @@ local function update_month(ts, cid, day_changed)
         current_month = month
     end
 
-    local country = fx.get_country_id(read_message("Fields[geo]"))
+    local country = fx.get_country_id(read_message("Fields[country]"))
```
mreid-moz, line 94: Do we need to `s/geo/country/` on line 136 too? or is it called "geo" for compatibility?

heka/sandbox/filters/firefox_weekly_dashboard.lua
```diff
@@ -81,7 +81,7 @@ local function update_week(ts, cid, day)
         error("data is in the past, this report doesn't back fill")
     end
 
-    local country = fx.get_country_id(read_message("Fields[geo]"))
+    local country = fx.get_country_id(read_message("Fields[country]"))
```
mreid-moz, line 84: Do we need to `s/geo/country/` on line 121 too? or is it called "geo" for compatibility?

trink:
geo is the column header used in the current dashboard (so the output should be 100% compatible/expected).

mreid-moz:
Ok, let's keep it compatible.

PR #94, opened 2015-07-10
Increase the S3 output stall check to one minute (instead of checking…

trink:
… every

dashboard ticker_interval)

### DIFF
--- a/heka/sandbox/filters/telemetry_s3output_monitors.lua
+++ b/heka/sandbox/filters/telemetry_s3output_monitors.lua
@@ -25,10 +25,10 @@ local alert = require "alert"
 
 local plugins        = {}
 
-local function find_plugin(name)
+local function find_plugin(name, ts)
     local p = plugins[name]
     if not p then
-        p = {last_alert = 0, last_pff = 0, last_pmc = 0}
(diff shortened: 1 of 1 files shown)


heka/sandbox/filters/telemetry_s3output_monitors.lua
```diff
@@ -54,10 +56,16 @@ function process_message ()
             if v.Name ~= "TelemetryErrorsOutput" then
                 n = v.ProcessMessageCount.value
                 if type(n) == "number" then
-                    if n == p.last_pmc then -- no message in the Dashboard ticker_interval
-                        p.msg = string.format("%s ProcessMessageCount has stalled at %d", v.Name, n)
+                    if n == p.last_pmc then
+                        if p.last_update + 60 * 1e9 < ts then
```
mreid-moz, line 60: Should we make the "60" here a config setting?

trink, line 60: If we ran multiple instances with different settings then a config option (specified in puppet) would be my choice otherwise I would rather modify the source.

mreid-moz, line 60: Ok, let's leave it until we need to run multiples.

PR #97, opened 2015-07-15
Update OS Bucketing, normalize crash data.

mreid-moz:
Crash data uses a value of "Windows_NT" for the "os" field. Also,
channel and os are already normalized in the executive summary, but
not in crashes, so we need to normalize before getting ids.

### DIFF
--- a/heka/sandbox/modules/fx.lua
+++ b/heka/sandbox/modules/fx.lua
@@ -27,7 +27,7 @@ local function anywhere (p)
 end
 
 local normalize_os_grammar =
-(l.P"Windows" + "WINNT" + "Windows_NT") / "Windows" +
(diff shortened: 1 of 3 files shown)


heka/sandbox/modules/fx.lua
```diff
@@ -27,7 +27,7 @@ local function anywhere (p)
 end
 
 local normalize_os_grammar =
-(l.P"Windows" + "WINNT") / "Windows" +
+(l.P"Windows" + "WINNT" + "Windows_NT") / "Windows" +
```
trink: "Windows" would have already matched it

trink:
Monthly will need the same change

PR #100, opened 2015-07-22
Add telemetryEnabled to elasticsearch

whd:
https://bugzilla.mozilla.org/show_bug.cgi?id=1184383

### DIFF
--- a/heka/sandbox/encoders/es_fields.lua
+++ b/heka/sandbox/encoders/es_fields.lua
@@ -98,6 +98,10 @@ local info_fields = {
   , "subsessionLength"
 }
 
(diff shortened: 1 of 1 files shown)


heka/sandbox/encoders/es_fields.lua
```diff
@@ -156,6 +160,15 @@ function process_message()
         end
     end
 
+    local ok, info = pcall(cjson.decode, read_message("Fields[environment.settings]"))
+    if ok then
+        for i, k in ipairs(environment_fields) do
```
trink, line 165: The loop seems a little over kill but I assume you expect to be adding more values

PR #101, opened 2015-07-22
Propagate landfill messages for consolidated edge multi-decoder

whd:


heka/sandbox/decoders/extract_executive_summary.lua
```diff
@@ -137,7 +137,12 @@ local msg = {
 
 
 function process_message()
-    if read_message("Type") ~= "telemetry" then return 0 end
+    local typ = read_message("Type")
+    if typ == "heka.httpdata.request" and duplicate_original then
+        inject_message(read_message("raw"))
+        return 0
+    end
+    if typ ~= "telemetry" then return 0 end
```
trink: This change is unnecessary, the return 0 will keep the message.

whd: Removed.

PR #103, opened 2015-07-24
Add app, version, and vendor to the executive summary

trink:


heka/sandbox/decoders/extract_executive_summary.lua
```diff
@@ -175,6 +181,24 @@ function process_message()
         msg.Fields.buildId.value = bid
     end
 
+    msg.Fields.app.value = ""
```
mreid-moz: Can we reduce the boilerplate here?  Maybe a function like:
`set_string_value(msg.Fields.app, read_message("Fields[appName]"))`
that encapsulates this repeated logic?

PR #107, opened 2015-08-03
Add breakouts for Windows OS segments

trink:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1189863

### DIFF
--- a/heka/sandbox/modules/fx.lua
+++ b/heka/sandbox/modules/fx.lua
@@ -195,13 +195,13 @@ end
 function get_os_win_id(version)
     local ver = tonumber(version) or 0
 
-    local id = 0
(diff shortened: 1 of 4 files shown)


heka/sandbox/modules/fx.lua
```diff
@@ -168,6 +192,21 @@ function get_os_id(name)
     return 0
 end
 
+function get_os_win_id(version)
+    local ver = tonumber(version) or 0
+
+    local id = 0
+    if ver >= 10 and ver < 11 then
+        id = 1
```
mreid-moz: This would be clearer and easier to modify later as `id = os_win_ids["Windows 10"]` (same for line 202 and 204)

PR #111, opened 2015-08-11
Add os, sessionLength, and architecture to elasticsearch

whd:
`os` is a top-level field and the config for that is in puppet.


heka/sandbox/encoders/es_fields.lua
```diff
@@ -169,6 +170,11 @@ function process_message()
         end
     end
 
+    local ok, payload = pcall(cjson.decode, read_message("Payload"))
+    if ok and payload.application.architecture then
```
trink: Check the type on payload.application.  If it is not a table this encoder will be terminated.

whd:
Added the check.

PR #112, opened 2015-08-14
Add a subsessionTimestamp to the executive summary

trink:


heka/sandbox/decoders/extract_executive_summary.lua
```diff
@@ -232,6 +237,20 @@ function process_message()
             if type(info.subsessionCounter) == "number" then
                 msg.Fields.subsessionCounter.value = info.subsessionCounter
             end
+
+            if type(info.subsessionStartDate) == "string" then
+                local t = dt.rfc3339:match(info.subsessionStartDate)
```
trink, line 228: This timestamp is a little screwed up (some values aren't ISO as spec'ed). I am thinking we may just want the date part (treat it UTC and ignore the timezone). i.e. just keep the else clause.

`2015-08-05T00:00:00.0-07:00`
`2015-08-05T00:00:00`

heka/sandbox/decoders/extract_executive_summary.lua
```diff
@@ -232,6 +237,20 @@ function process_message()
...
+                if t then
+                    msg.Fields.subsessionTimestamp.value = dt.time_to_ns(t)
+                else -- some dates are not RFC compliant
+                    t = broken_date:match(info.subsessionStartDate)
+                    if t then
+                        msg.Fields.subsessionTimestamp.value = dt.time_to_ns(t)
+                    else
+                        msg.Fields.subsessionTimestamp.value = cts
```
mreid-moz: I don't think we should fall back to creationTimestamp here - we already have that value in a separate field, so we should just leave out the subsessionTimestamp if it's unavailable (or unparseable).

trink: The filtering and computation work off of subsessionTimestamp.  If it is not set the code will have to fall back to creationTimestamp anyway.  I can remove it from here but would have to push it out to any consumers of the data.

Edit: would still have to special case crash pings unless I use the crash date for the subsessionTime

mreid-moz: Yeah, I'd rather let consumers decide what to do about it.

trink:
Cleaning up the executive_summary as discussed in IRC
- removing un-used sessionId and subsessionCounter
- standardizing on an activityTimestamp

heka/sandbox/decoders/extract_executive_summary.lua
```diff
@@ -211,26 +212,28 @@ function process_message()
...
         set_string_field(msg.Fields.reason, "Fields[reason]")
-        msg.Fields.hours.value              = 0
-        msg.Fields.google.value             = 0
-        msg.Fields.bing.value               = 0
-        msg.Fields.yahoo.value              = 0
-        msg.Fields.other.value              = 0
-        msg.Fields.sessionId.value          = ""
-        msg.Fields.subsessionCounter.value  = 0
-        msg.Fields.pluginHangs.value        = 0
+        msg.Fields.hours.value               = 0
```
mreid-moz: nit: some of this alignment whitespace can be removed (also above in "crash_fields" and "main_fields").

PR #113, opened 2015-08-19
[do not merge] Make a branch that can build kafka consumer group code

whd:
There is a long story here involving the various heka, data-pipeline, snappy, and sarama repositories, but the short version is that in order to build this consumer group code without going into dependency hell I had to remove the other kafka plugins from the heka branch referenced in this PR. The actual code is here: https://github.com/mozilla-services/heka/compare/dev...whd:kafka_consumer_group.

This is being tested in stage, but preliminary tests look good: multiple consumers leaving/joining a consumer group works as expected.


whd:
See #125.

PR #117, opened 2015-09-08
Add a filter to summarize data volume.

mreid-moz:
Step one in Bug 1179751 - summarize data volume by channel, build
date, and submission date.

### DIFF
--- a/heka/sandbox/filters/dollars.lua
+++ b/heka/sandbox/filters/dollars.lua
@@ -23,6 +23,9 @@ factors relevant to budget planning. See Bug 1179751.
 
 require "os"
 require "table"
+
+local bdate_grammar
+
 if not TestMode then
     require "lpeg"
     require "cjson"
@@ -45,9 +48,9 @@ local counters = {
     build = {},
 }
 
-function get_build_date(buildid)
-    if not buildid then
-        return buildid
+local function get_build_date(buildid)
+    if type(buildid) ~= "string" then
+        return nil
     end
 
     if string.len(buildid) < 8 then
@@ -62,12 +65,12 @@ function get_build_date(buildid)
 end
 
 -- If possible, evict a key older than "newdate" to make room for the new key.
-function evict(counter, newdate)
+local function evict(counter, newdate)
     local evictable_keys = {}
     local count = 0
     for k, v in pairs(counter) do
         count = count + 1
-        if v ~= nil and k < newdate then
+        if k < newdate then
             table.insert(evictable_keys, k)
         end
     end
@@ -85,9 +88,9 @@ function evict(counter, newdate)
     return to_delete
 end
 
-function record_size(sizes, counter, channel, date, docType, size, count)
+local function record_size(sizes, counter, channel, date, docType, size, count)
     -- By default, count 1.
-    if not count then
+    if type(count) ~= "number" then
         count = 1
     end
 
@@ -97,7 +100,8 @@ function record_size(sizes, counter, channel, date, docType, size, count)
     end
     local sc = sizes[channel]
 
-    if not sc[date] then
+    local scd = sc[date]
+    if not scd then
         if counter[channel] >= max_per_channel then
             removed = evict(sc, date)
             if removed == 0 then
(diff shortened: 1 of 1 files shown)


heka/sandbox/filters/dollars.lua
```diff
@@ -0,0 +1,268 @@
...
+    [PipelineBudget]
+    type = "SandboxFilter"
+    filename = "lua_filters/dollars.lua"
+    message_matcher = "Type == 'telemetry'"
+    ticker_interval = 60
+    preserve_data = false
+
+--]]
+
+require "os"
```
trink, line 24: Most of the functions don't need to be at global scope, same for the bdate_grammar

heka/sandbox/filters/dollars.lua
```diff
@@ -0,0 +1,268 @@
...
+}
+
+-- Key count by channel (for eviction purposes)
+local counters = {
+    submission = {},
+    build = {},
+}
+
+function get_build_date(buildid)
+    if not buildid then
```
trink: test for string type, not existence

heka/sandbox/filters/dollars.lua
```diff
@@ -0,0 +1,268 @@
...
+    return string.sub(buildid, 1, 8)
+end
+
+-- If possible, evict a key older than "newdate" to make room for the new key.
+function evict(counter, newdate)
+    local evictable_keys = {}
+    local count = 0
+    for k, v in pairs(counter) do
+        count = count + 1
+        if v ~= nil and k < newdate then
```
trink: v cannot be nil

heka/sandbox/filters/dollars.lua
```diff
@@ -0,0 +1,268 @@
...
+        to_delete = to_delete + 1
+        counter[evictable_keys[to_delete]] = nil
+        count = count - 1
+    end
+    return to_delete
+end
+
+function record_size(sizes, counter, channel, date, docType, size, count)
+    -- By default, count 1.
+    if not count then
```
trink: again test the type, not existence

heka/sandbox/filters/dollars.lua
```diff
@@ -0,0 +1,268 @@
...
+            if removed == 0 then
+                -- This date is too old and our table is full.
+                return
+            end
+            counter[channel] = counter[channel] - removed
+        end
+        counter[channel] = counter[channel] + 1
+        sc[date] = {}
+    end
+    local scd = sc[date]
```
trink: this could be assigned before the if (instead of always looking it up twice) and explicitly set to the new table in the failed case.

heka/sandbox/filters/dollars.lua
```diff
@@ -0,0 +1,268 @@
...
+
+    if not scd[docType] then
+        scd[docType] = {}
+    end
+    local scdt = scd[docType]
+
+    scdt.count = (scdt.count or 0) + count
```
trink: count and size should be initialized in the table constructor

heka/sandbox/filters/dollars.lua
mreid-moz: setting a value to `nil` is the way I remove it from the table - does that mean that `pairs` automatically skips anything with a nil value?

trink: yes

heka/sandbox/filters/dollars.lua
```diff
@@ -0,0 +1,272 @@
...
+        end
+        counter[channel] = counter[channel] + 1
+        scd = {}
+        sc[date] = scd
+    end
+
+    if not scd[docType] then
+        scd[docType] = {count = 0, size = 0}
+    end
+    local scdt = scd[docType]
```
trink, line 113: could also move this one above the if to avoid the double lookup

heka/sandbox/filters/dollars.lua
```diff
@@ -0,0 +1,271 @@
...
+    while count >= max_per_channel do
+        -- wipe out the min key.
+        to_delete = to_delete + 1
+        counter[evictable_keys[to_delete]] = nil
+        count = count - 1
+    end
+    return to_delete
+end
+
+local function record_size(sizes, counter, channel, date, docType, size, count)
```
trink: count is only used in one unit test is this argument needed?

PR #118, opened 2015-09-09
Decoder view config

whd:
Make the decoder match configurable and skip non-telemetry decoders (we have nginx decoders on the edge now).

### DIFF
--- a/heka/sandbox/filters/telemetry_decoder_view.lua
+++ b/heka/sandbox/filters/telemetry_decoder_view.lua
@@ -81,18 +81,16 @@ function process_message ()
             end
 
             local id = string.match(v.Name, decoder_match)
-            if not id then
-                return -1, "Telemetry decoder is missing its identifier"
-            else
+            if id then
                 id = tonumber(id)
-            end
 
-            if type(v["ProcessMessageCount-TelemetryDecoder"]) == "table" then
-                update_delta(ts, TOTAL, id, id_count, v["ProcessMessageCount-TelemetryDecoder"].value)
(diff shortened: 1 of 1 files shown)


heka/sandbox/filters/telemetry_decoder_view.lua
```diff
@@ -79,19 +80,17 @@ function process_message ()
...
+                if type(v["ProcessMessageCount-TelemetryDecoder"]) == "table" then
+                    update_delta(ts, TOTAL, id, id_count, v["ProcessMessageCount-TelemetryDecoder"].value)
+                end
 
-            if type(v["ProcessMessageFailures-TelemetryDecoder"]) == "table" then
-                update_delta(ts, FAILURES, id, id_failures, v["ProcessMessageFailures-TelemetryDecoder"].value)
+                if type(v["ProcessMessageFailures-TelemetryDecoder"]) == "table" then
+                    update_delta(ts, FAILURES, id, id_failures, v["ProcessMessageFailures-TelemetryDecoder"].value)
+                end
             end
```
mreid-moz, line 94: Do you not care about the case where "id" is missing now? Or should it still have an "else return -1, ..." bit here?

whd, line 94: Yes, since there is only one match string, Previously the error message "id" is missing was misleading in some cases because the other decoder name didn't match the pattern at all. We could make it more explicit with two matches (one for the prefix and one for the id), but I don't think it's worth it.

PR #119, opened 2015-09-10
Try to fix the travis and osx builds for GeoIP

mreid-moz:
Recent changes require at least version 1.6.3 of the geoip library .

PR #120, opened 2015-09-10
Don't run make test, as it is broken upstream

whd:
```
[12:49:13] <RaFromBRC> unfortunately, we had to turn off heka travis b/c it was consistently failing (not always in the same ways, mind you) in ways we were unable to reproduce outside of travis :P
```


mreid-moz:
Build still fails due to geoip library being too old. I've rolled this into changes over in PR #119

PR #126, opened 2015-09-17
Remove useless assignments.

mreid-moz:
Heka overwrites Type and Logger automatically. Type is replaced with
"heka.sandbox.stem", where "stem" is the filename of the lua filter
without the .lua file extension. Logger is replaced with the plugin
name.

So assigning to these two fields in the lua code has no effect.


whd:
Type is actually prepended with “heka.sandbox.” by the SandboxFilter to avoid data confusion/mis-representation
[(source)](http://hekad.readthedocs.org/en/v0.10.0b1/sandbox/index.html). I filed #132 to add it back.

PR #134, opened 2015-09-23
Add the profile creation date to the executive summary

trink:
- update the executive report to use the profile creation date when
  setting the new user flag

### DIFF
--- a/heka/sandbox/decoders/extract_executive_summary.lua
+++ b/heka/sandbox/decoders/extract_executive_summary.lua
@@ -110,6 +110,19 @@ local function get_os_version()
 end
 
 
+local function get_profile_creation()
+    local json = read_message("Fields[environment.profile]")
+    local ok, profile = pcall(cjson.decode, json)
+    if not ok then return 0 end
+
+    local days = profile.creationDate
+    if type(days) == "number" and days > 0 then
+        return days * SEC_IN_DAY * 1e9
+    end
+    return 0
+end
+
+
 local function set_string_field(field, key)
     field.value = ""
     local s = read_message(key)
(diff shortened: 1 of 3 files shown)


heka/sandbox/decoders/extract_executive_summary.lua
```diff
@@ -110,6 +110,19 @@ local function get_os_version()
...
 
 
+local function get_profile_creation()
+    local json = read_message("Fields[environment.profile]")
+    local ok, profile = pcall(cjson.decode, json)
+    if not ok then return 0 end
+
+    local days = profile.creationDate
+    if type(days) == "number" and days > 0 then
+        return days * SEC_IN_DAY * 1e9
```
mreid-moz, line 120: Do we need to check if it's in the future?

trink, line 120: Not really, it will just count as new once

heka/sandbox/filters/firefox_executive_report.lua
```diff
@@ -85,7 +86,10 @@ end
 
 local function update_row(r, cid, country, channel, _os, dow)
     local dflt = fx.get_boolean_value(read_message("Fields[default]"))
-    fx_cids:add(cid, country, channel, _os, dow, dflt)
+    local pct = read_message("Fields[profileCreationTimestamp")
```
mreid-moz: Missing trailing bracket in after `profileCreationTimestamp`

heka/sandbox/decoders/extract_executive_summary.lua
mreid-moz, line 120: Ah, I see it gets reset [here](https://github.com/mozilla-services/data-pipeline/blob/profile_creation/heka/plugins/fx/executive_report.c#L396). I thought it would report as "new" once per period. Never mind.

heka/sandbox/decoders/extract_executive_summary.lua
```diff
@@ -110,6 +110,19 @@ local function get_os_version()
...
+    end
+    return 0
```
mreid-moz, line 122: I think we should default to "today" as the creation date if it is missing. But that's just my opinion, we should confirm with the metrics team.

mreid-moz:
Let's go ahead with defaulting to "not new" if profile creation date is missing, then update the code again if needed.

mreid-moz:
FWIW, I checked with bcolloran and joy and they say we should not default to new, so I think we're good.

PR #135, opened 2015-09-28
Bugzilla 1204589 - move raw bytes out of the msg.Payload string

trink:


whd:
This looks good to me, and has been tested in stage.

PR #136, opened 2015-10-01
Check field value is not nil instead of its direct value

whd:
This should fix `telemetryEnabled` being unset (and converted by the s3 schema to `UNKNOWN`) instead of `false`.

### DIFF
--- a/heka/sandbox/decoders/extract_telemetry_dimensions.lua
+++ b/heka/sandbox/decoders/extract_telemetry_dimensions.lua
@@ -46,7 +46,7 @@ local duplicate_original = read_config("duplicate_original")
 -- telemetry messages should not contain duplicate keys so this function
 -- replaces/removes the first key that exists or adds a new key to the end
 local function update_field(fields, name, value)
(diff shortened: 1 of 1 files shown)


trink:
this is an improvement but it doesn't prevent it from being a non boolean value.

whd:
At this point we want to allow non-boolean values (the client occasionally sends "false" instead of false, at least in the specific case of `telemetryEnabled`). The s3 schema logic automatically converts the data to a string for the purposes of schema validation (we have tests for this) so I think it's fine to leave as-is. Lua code using `telemetryEnabled` would need to type check and do extra validation, but I don't think we will have any lua consumers of this field.

trink:
`the client occasionally sends "false"`
I would propose any crap like that we normalize here so other consumers don't have to check if it is "false", "False", false, 0, etc.

whd:
I would like this piece merged prior to the application of any normalization logic, for which we should probably open a bug and do an RFC. We should probably open a separate bug anyway for why the client is sending mixed types in the first place.

PR #140, opened 2015-10-15
Remove the original data on successful telemetry dimension extraction

trink:


mreid-moz:
Looks good, just waiting for the build to pass.

mreid-moz:
@whd we should deploy this asap.  Not sure if we need to clean up prod data that still has the "submission" field - I guess that depends on whether we can fix the spark reader code to stop blowing up when treating the bytes as JSON.

PR #145, opened 2015-11-03
Add a derived streams module

trink:
- split out common output formats
- add a fx.ping library with caching and common functionality
- add derived streams for the main and crash telemetry pings
- update the build to pull in the lua postgres module
- fix output and io_modules installation


hindsight/io_modules/derived_stream.lua
```diff
@@ -0,0 +1,197 @@
...
+            process_message =  function ()
+                local fh = get_output_fh(read_message(ts_field), "tsv")
+                tsv.write_message(fh, schema, nil_value)
+                return 0
+            end
+        end
+
+        return process_message, timer_event
+    else
+        error("invalid derived stream format: " .. format)
```
rafrombrc: I haven't done a review of this code, but while testing it I discovered that omitting the `format` setting will mean the format var is nil and this concatenation will fail. Should either use `tostring(format)` in the concatenation, or explicitly check for format being nil.

trink: Thanks, I threw a tostring around it (it will be committed with a few other changes)

hindsight/io_modules/derived_stream/redshift.lua
```diff
@@ -0,0 +1,95 @@
...
+    return tostring(v)
+end
+
+function esc_str(con, v, max)
+    if v == nil then return "NULL" end
+    if max == nil then max = VARCHAR_MAX_LENGTH end
+    if type(v) ~= "string" then v = tostring(v) end
+    if string.len(v) > max then v = string.sub(v, 1, max) end
+
+    local escd = con:escape(v)
```
trink: escape() can return a string that fails to load into the DB
PostgreSQL: ERROR:  CHAR string contains invalid ASCII character: c2 a

hindsight/output/crash_summary.lua
```diff
@@ -0,0 +1,73 @@
...
+local ds = require "derived_stream"
+local fx = require "fx"
+local ping = require "fx.ping"
+
+local name = "crash_summary"
+local schema = {
+--  column name                 type            length  attributes  field /function
+    {"Timestamp"                ,"TIMESTAMP"    ,nil    ,"SORTKEY"  ,"Timestamp"},
+    {"crashDate"                ,"DATE"         ,nil    ,nil        ,function () return ping.get_date(ping.payload().payload.crashDate) end},
+    {"clientId "                ,"CHAR"         ,36     ,"DISTKEY"  ,"Fields[clientId]"},
```
trink: remove extra space

hindsight/output/executive_summary.lua
```diff
@@ -0,0 +1,117 @@
...
+    {"os"                           ,"VARCHAR"      ,7      ,nil        ,function () return fx.normalize_os(read_message("Fields[os]")) end},
+    {"osVersion"                    ,"VARCHAR"      ,32     ,nil        ,function () return ping.system().os.version end},
+    {"app"                          ,"VARCHAR"      ,32     ,nil        ,"Fields[appName]"},
+    {"version"                      ,"VARCHAR"      ,32     ,nil        ,"Fields[appVersion]"},
+    {"vendor"                       ,"VARCHAR"      ,32     ,nil        ,"Fields[appVendor]"},
+    {"reason"                       ,"VARCHAR"      ,32     ,nil        ,"Fields[reason]"},
+    {'"default"'                    ,"BOOLEAN"      ,nil    ,nil        ,ping.is_default_browser},
+    {"hours"                        ,"FLOAT8"       ,nil    ,nil        ,ping.hours},
+    {"google"                       ,"INT"          ,nil    ,nil        ,function () return search_counts[1] end},
+    {"bing  "                       ,"INT"          ,nil    ,nil        ,function () return search_counts[2] end},
```
trink: remove extra spaces

hindsight/output/main_summary.lua
```diff
@@ -0,0 +1,90 @@
...
+local ds = require "derived_stream"
+local fx = require "fx"
+local ping = require "fx.ping"
+
+local name = "main_summary"
+local schema = {
+--  column name                     type            length  attributes  field /function
+    {"Timestamp"                    ,"TIMESTAMP"    ,nil    ,"SORTKEY"  ,"Timestamp"},
+    {"subsessionDate"               ,"DATE"         ,nil    ,nil        ,function () return ping.get_date(ping.info().subsessionStartDate) end},
+    {"clientId "                    ,"CHAR"         ,36     ,"DISTKEY"  ,"Fields[clientId]"},
```
trink: remove space

hindsight/io_modules/derived_stream.lua
```diff
@@ -0,0 +1,197 @@
...
+                end
+            end
+            return -4
+        end
+
+        local function timer_event(ns, shutdown)
+            if shutdown or last_insert + ticker_interval <= ns / 1e9 then
+                local err = insert_files()
+                if not err then
+                    batch_checkpoint_update()
```
rafrombrc: This is causing an error on shutdown for me: `attempt to call upvalue 'batch_checkpoint_update' (a nil value)`.

trink: You need the latest Hindsight/lua_sandbox to run this module

trink:
Don't merge yet.  This PR will hopefully be redesigned, closed, squashed, and re-submitted due to the experience on the c3.8xlarge.

hindsight/io_modules/derived_stream/redshift.lua
mreid-moz: What encoding was the database created with? We should probably use UTF8 and not ASCII.

hindsight/io_modules/derived_stream.lua
```diff
@@ -0,0 +1,197 @@
...
+setfenv(1, M) -- Remove external access to contain everything in the module
+
+local SEC_IN_DAY = 60 * 60 * 24
+
+function load_schema(name, schema)
+    local ticker_interval   = read_config("ticker_interval")
+    local cfg_name          = read_config("cfg_name")
+    local format            = read_config("format")
+    local ts_field          = read_config("ts_field") or "Timestamp"
+    local files             = {} -- manages the derive stream buffer/output files
```
mreid-moz: s/derive/derived/

hindsight/io_modules/derived_stream.lua
```diff
@@ -0,0 +1,197 @@
...
+
+        local function get_output_file(ts)
+            local day = math.floor(ts / (SEC_IN_DAY * 1e9))
+            local f = files[day]
+            if not f then
+                local date = os.date("%Y%m%d", ts / 1e9)
+                local table_name = string.format("%s_%s", name, date)
+                local filename = string.format("%s/%s_%s.sql", buffer_path, cfg_name, date)
+                local ok, cnt, err = pcall(con.execute, con, rs.get_create_table_sql(table_name, schema))
+                if ok and err and not string.match(err, "duplicate key violates unique constraint") then
```
mreid-moz, line 144: Please add a comment that the string match detects concurrent "create table" statements, and that failure is expected.

hindsight/io_modules/derived_stream.lua
```diff
@@ -0,0 +1,197 @@
...
+                if file.offset == 0 then
+                    file.fh:write("INSERT INTO ", file.table_name, " VALUES ")
+                else
+                    file.fh:write(",")
+                end
+                rs.write_values_sql(file.fh, con, schema)
+                file.offset = file.fh:seek("end")
+            end
+
+            if not file or file.offset >= buffer_size then
```
rafrombrc, line 196: I'm hitting a case where `file` is not nil but `file.offset` is nil, causing this comparison to fail with `attempt to compare number with nil`.

rafrombrc, line 196: Yup, verified that the `file.fh:seek("end")` on L102 returns `nil` in error cases, which falls through to cause the compare on L105 to fail.

trink, line 196: What is the error case were you have a valid file handle but seek fails?

I think the only time I have seen that failure is after running out of disk space

rafrombrc, line 196: Yup, that's what happened, "No space left on device." Probably not too important to deal with here, considering if that's the case we have bigger problems...

trink, line 196: Yeah, we would still error and exit so the question is do you want a nicer error message?

rafrombrc, line 196: Might as well, might save some debugging work.

hindsight/io_modules/derived_stream.lua
```diff
@@ -0,0 +1,197 @@
...
+                    os.remove(v.filename);
+                end
+            end
+            last_insert = os.time()
+        end
+
+        local function process_message()
+            local file
+            if not (uuid and uuid == read_message("Uuid")) then -- make sure we aren't in a retry loop
+                file, err = get_output_file(read_message(ts_field))
```
rafrombrc, line 182: Seems like `err` needs to be declared as local here, yes?

trink, line 182: Yup, added

hindsight/io_modules/derived_stream.lua
```diff
@@ -0,0 +1,197 @@
...
+            end
+            return f
+        end
+
+        local function insert_files() -- all files are flushed at once to ensure the checkpoint is accurate
+            for k,v in pairs(files) do
+                if v.fh and v.offset ~= 0 then
+                    v.fh:seek("set")
+                    local ok, cnt, err = pcall(con.execute, con, v.fh:read("*a")) -- read the entire file and execute the query
+                    if ok and err then -- database error
```
trink: Handle lost connections

hindsight/io_modules/derived_stream.lua
```diff
@@ -0,0 +1,197 @@
...
+        local db_config     = read_config("db_config") or error("db_config must be set")
+        local buffer_path   = read_config("buffer_path") or error("buffer_path must be set")
+        local buffer_size   = read_config("buffer_size") or 10000 * 1024
+        assert(buffer_size > 0, "buffer_size must be greater than zero")
+
+        local last_insert   = 0
+        local uuid          = nil
+
+        local env = assert(driver.postgres())
+        local con = assert(env:connect(db_config.name, db_config.user, db_config._password, db_config.host, db_config.port))
```
trink: Allow keyword/value connection strings to provide more control i.e. timeouts?
http://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-CONNSTRING

trink:
I won't re-write the history but a clean PR with some additional fixes will be opened tomorrow.

PR #151, opened 2015-11-12
Reopen the geoip DB once an hour

whd:


mreid-moz:
Looks good to me.

heka/sandbox/decoders/http_edge_decoder.lua
```diff
@@ -99,6 +102,13 @@ function process_message()
     landfill_msg.Fields.submissionDate = os.date("%Y%m%d", landfill_msg.Timestamp / 1e9)
     landfill_msg.Fields.submission = {value = read_message("Payload"), value_type = 1} -- Bugzilla 1204589
 
+    -- Reopen city_db once an hour.
+    local current_hour = math.floor(os.time() / 3600)
+    if current_hour > hour then
+        city_db:close()
+        city_db = assert(geoip.city.open(read_config("geoip_city_db")))
```
trink, line 109: How does the new install happen?  i.e., will it always be available at this point? If this open fails the decoder is terminated.

whd, line 109: This happens via yum, 5 minutes before the hour. There is a _very_ slight chance that the network update is so slow that the file is being updated while lua is reloading the db, but in that case systemd will restart the service anyway.

Issue #147, opened 2015-11-06
Investigate why heka-s3cat cannot saturate the network interface


trink:
Even on a 1Gb/s system

PR #153, opened 2015-11-13
Issue #147 Utilize all the network bandwith when pulling data from S3

trink:


heka/cmd/s3cat/main.go
```diff
@@ -0,0 +1,146 @@
...
+	for attempt := 1; attempt <= 5; attempt++ {
+		rc, err := getS3Reader(bucket, s3Key, lastGoodOffset)
+		if err != nil && err != io.EOF {
+			fmt.Fprintf(os.Stderr, "Error in attempt %d reading %s at offset %d: %s\n", attempt, s3Key, lastGoodOffset, err)
+			continue RetryS3
+		} else {
+			nr := bufio.NewReader(rc)
+			n, err := nr.WriteTo(os.Stdout)
+			if err != nil && err != io.EOF {
+				fmt.Fprintf(os.Stderr, "Error in attempt %d writing %s at offset %d: %s\n", attempt, s3Key, lastGoodOffset, err)
```
trink, line 117: The only time I have seen this fail is with a broken pipe... retry won't help so we should probably just exit.

mreid-moz, line 117: +1 on aborting on failed write.  It's unlikely to happen from something recoverable.

trink, line 117: Now seeing other errors too "connection reset by peer" which we do want to retry
I will do some more testing and special case the broken pipe.

mreid-moz, line 117: on Write? We definitely want to retry the Read....

mreid-moz, line 117: Ok, if it can throw read or write errors, you should probably leave the retry :-/

PR #154, opened 2015-11-13
Add an optional 'crashes' field.

mreid-moz:
Add a field 'crashes' which optionally contains the count of crashes.
This allows several crashes to be represented by a single record. If
the field is missing, we count one crash, same as before this change.

### DIFF
--- a/heka/sandbox/filters/firefox_executive_report.lua
+++ b/heka/sandbox/filters/firefox_executive_report.lua
@@ -99,7 +99,17 @@ local function update_row(r, cid, country, channel, _os, dow)
         r[12] = r[12] + (tonumber(read_message("Fields[yahoo]")) or 0)
         r[13] = r[13] + (tonumber(read_message("Fields[other]")) or 0)
     elseif doc_type == "crash" then
-        r[8] = r[8] + 1
+        local num_crashes = read_message("Fields[crashes]")
+        -- If field is missing, assume one crash per record.
+        if num_crashes == nil then
+            num_crashes = 1
+        end
+
+        -- If field contains an unexpected value, assume zero crashes.
+        if type(num_crashes) ~= "number" or num_crashes < 0 then
+            num_crashes = 0
+        end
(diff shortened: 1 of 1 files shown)


mreid-moz:
Note to self: this will require an update to the Executive Summary Schema documentation.

r? @trink - does this seem reasonable?

heka/sandbox/filters/firefox_executive_report.lua
```diff
@@ -99,7 +99,17 @@ local function update_row(r, cid, country, channel, _os, dow)
         r[12] = r[12] + (tonumber(read_message("Fields[yahoo]")) or 0)
         r[13] = r[13] + (tonumber(read_message("Fields[other]")) or 0)
     elseif doc_type == "crash" then
-        r[8] = r[8] + 1
+        local num_crashes = read_message("Fields[crashes]")
+        -- If field is missing, assume one crash per record.
+        if num_crashes == nil then
```
trink, line 104: just test the type here

heka/sandbox/filters/firefox_executive_report.lua
```diff
@@ -99,7 +99,17 @@ local function update_row(r, cid, country, channel, _os, dow)
...
+            num_crashes = 1
+        end
+
+        -- If field contains an unexpected value, assume zero crashes.
+        if type(num_crashes) ~= "number" or num_crashes < 0 then
+            num_crashes = 0
```
trink, line 110: Counting less than zero as zero here feels weird since a crash is specifically being logged but I am not sure if 1 makes any more sense (crap like this should go in the error bucket)

mreid-moz, line 110: counting it as zero basically amounts to skipping the record, which seems like the right thing to do on present-but-unexpected values.

trink, line 110: ok, then ignore both comments.

r+

PR #159, opened 2015-11-19
Use pg_config to find include dir if possible.

### LINKED ISSUE #157: Use pg_config to determine the include path
In the build script, we should set the postgres include path using something like `$(pg_config --includedir)`

mreid-moz:
Fixes #157


trink:
+1 interim fix

PR #160, opened 2015-11-19
Escape NUL character in the psv

trink:
- Redshift treats them as an end of record (EOR) marker, causing the
  remainder of the record to be truncated

### DIFF
--- a/hindsight/io_modules/derived_stream/redshift/psv.lua
+++ b/hindsight/io_modules/derived_stream/redshift/psv.lua
@@ -13,13 +13,15 @@ local string    = require "string"
 
 setfenv(1, M) -- Remove external access to contain everything in the module
 
-local esc_chars = { ["|"] = "\\|", ["\r"] = "\\r", ["\n"] = "\\n", ["\0"] = "0" }
+local esc_chars = { ["|"] = "\\|", ["\r"] = "\\r", ["\n"] = "\\n" }
 function esc_varchar(v, max)
     if v == nil then return "" end
     if max == nil then max = rs.VARCHAR_MAX_LENGTH end
     if type(v) ~= "string" then v = tostring(v) end
     if string.len(v) > max then v = string.sub(v, 1, max) end
-    return string.gsub(v, "[|\r\n%z]", esc_chars)
+    local s, e = string.find(v, "%z")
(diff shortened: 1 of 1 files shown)


hindsight/io_modules/derived_stream/redshift/psv.lua
```diff
@@ -13,13 +13,13 @@ local string    = require "string"
 
 setfenv(1, M) -- Remove external access to contain everything in the module
 
-local esc_chars = { ["|"] = "\\|", ["\r"] = "\\r", ["\n"] = "\\n" }
+local esc_chars = { ["|"] = "\\|", ["\r"] = "\\r", ["\n"] = "\\n", ["\0"] = "0" }
```
trink: @whd @mreid do you want it removed from the string or should we just discard the entire string as garbage

Edit: the zero was mainly for debugging (so I can see what the garbarge string looked like e.g., `shutdown0000000000000000]000\\n000`

whd: I am for discarding it as garbage, but not strongly.

mreid-moz: Can we replace it with a literal `\0`? (2 chars, backslash, then zero).  It'd be nice to distinguish these from legitimate zeroes.

I would also be OK with truncating the string at the null char, or discarding the string as garbage (as long as we log some of them so we can debug and validate upstream if needed).

trink: Yeah, the thing that sucks about the entire string discard check is that it requires yet another scan over every string but replacing the bad character with 0 \0 or "" isn't super useful and truncating at the first NUL just hides the problem.

Edit: as for the log... you see the bad string above what is actually going to be done about it?... so we will create more data we actually don't use ;(  Most of this should be pushed into the data validation bugs.

mreid-moz: re: log, just a simple "last N observed strings containing NUL" heka plugin would do for me.  Just want some visibility into the problem.

trink:
In the end treating it like a c string seemed the best (for this specific type of error it produces the correct result).

PR #163, opened 2015-11-25
Fix the broken redshift/s3 insert/copy retry logic

trink:
- preserve the file object for the retry
- attempt 1 retry on shutdown and report error on failure
- range check the timestamp to prevent conversion failures


mreid-moz:
Only suggestion is to roll the "retry once" logic into the `insert_file` and `copy_file` functions.  Otherwise, looks good.

mreid-moz:
I'm going to merge it as-is, and filed #164 to follow up on the possible refactoring of retry logic.

PR #165, opened 2015-11-30
Verify the flash version is a string

trink:


hindsight/modules/fx/ping.lua
```diff
@@ -85,6 +85,7 @@ function flash_version()
     if type(addons.activePlugins) == "table" then
         for i,v in ipairs(addons.activePlugins) do
             if type(v) == "table" and v.name == "Shockwave Flash" then
+                if type(v.version) ~= "string" then v.version = "" end
```
mreid-moz, line 88: If it's a number, can (should?) we tostring it?

trink, line 88: No it will just fail to parse in that case also

mreid-moz, line 88: ok, makes sense.

PR #166, opened 2015-12-02
Update heka SHA to latest dev

whd:
This should definitely have the decoder race condition fix, and has been tested in stage.

/cc @rafrombrc


mreid-moz:
I don't think there's anything else I can add here, so just merging.

PR #168, opened 2015-12-02
List the 'telemetry-2' bucket.

mreid-moz:
Missed this during the review.


trink:
Some day we will have to fix it up to work with the combo of telemetry-2 and telemetry-release but for now this matches the current dashboard job.

PR #169, opened 2015-12-03
SQL-based implementation of the executive report.

mreid-moz:


reports/executive_summary/README.txt
```diff
@@ -5,13 +5,25 @@ To Deploy
...
     Output Directory:      output
     Output Visibility:     Private
-    Schedule Frequency:    Daily
-    Day of Week:           n/a (Sunday)
+    Schedule Frequency:    Weekly
+    Day of Week:           Monday
     Day of Month:          n/a (1)
-    Time of Day (UTC):     2pm
+    Time of Day (UTC):     10am
     Job Timeout (minutes): 120
```
trink: How quickly does the weekly job run?

reports/executive_summary/run.sh
```diff
@@ -1,42 +1,107 @@
...
+        --verbose \
+        --check-tables \
+        --db-url "$DB_URL" \
+        --report-start $TARGET \
+        --mode $MODE > "$CURRENT"
+
+OVERALL="v4-${MODE}.csv"
+DASHBOARD_S3="s3://net-mozaws-prod-metrics-data/firefox-executive-dashboard"
+echo "Fetching previous state from $OVERALL..."
+aws s3 cp "$DASHBOARD_S3/$OVERALL" ./
```
trink, line 80: We should check the return codes on all these network operations (I have had them fail before).  This could assume no previous state when there actually was one.

reports/executive_summary/README.txt
mreid-moz: I'm not sure with the change to fewer, larger machines in the cluster. I'll check again. Meanwhile, I'll bump this up to 300 for weekly and 600 for monthly just to be safe.

reports/executive_summary/run.sh
mreid-moz, line 80: Good idea. I will add that.

mreid-moz, line 80: Actually, I think some helper code for fetching state, metadata, credentials, etc could be in order...

PR #178, opened 2016-01-07
fix gnu/bsd incompatibilities.

miah:
OSX/BSD `sort` lacks `-V`, and `sed` lacks `-r`.

Still a remaining bug in that this only checks geoip version in brew and doesn't ensure that its actually installed.

### DIFF
--- a/bin/build_pipeline_heka.sh
+++ b/bin/build_pipeline_heka.sh
@@ -19,9 +19,9 @@ Darwin)
     echo "Ensuring we have the latest geoip for OSX..."
     MIN_GEOIP_VER=1.6.3
     if [ ! -z "$(which brew)" ]; then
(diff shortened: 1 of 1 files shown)


mreid-moz:
Ahh, I always use the GNU versions of coreutils on OSX.

We can't just replace `sort -V` with `sort`, see:

``` bash
$ cat t.txt
1.10.9
1.6.0
$ sort t.txt
1.10.9
1.6.0
$ sort -V t.txt
1.6.0
1.10.9
```

miah:
Ah ya, sort is tricky. I guess because the versions are similar enough I didn't run into that issue.

mreid-moz:
@miah did you find a way to sort version numbers with the osx/bsd `sort`?

georgf:
I addressed the sorting here:
https://github.com/georgf/data-pipeline/commit/1a05e89fcc270b4c01ddd179fe8013dc46c7d99e

@miah Do you want to push this here or should i open a new pull request?

PR #180, opened 2016-01-08
Add a filter to emit the webrtc derived stream

whd:
https://bugzilla.mozilla.org/show_bug.cgi?id=1231410

I'm using `read_message("raw")` when needed but not `decode_message` to decode the entire message always under the assumption that it's slightly more efficient to do it this way.

### DIFF
--- a/heka/sandbox/filters/telemetry_webrtc.lua
+++ b/heka/sandbox/filters/telemetry_webrtc.lua
@@ -21,9 +21,9 @@ Derived stream for webrtc. https://bugzilla.mozilla.org/show_bug.cgi?id=1231410
 require 'cjson'
 
 local function check_payload (payload)
-    local w = payload['webrtc'] or {}
-    local i = w['IceCandidatesStats'] or {}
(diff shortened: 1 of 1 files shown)


heka/sandbox/filters/telemetry_webrtc.lua
```diff
@@ -0,0 +1,59 @@
...
+function process_message()
+    local ok, json = pcall(cjson.decode, read_message("Payload"))
+    if not ok then return -1, json end
+    local p = json['payload'] or {}
+    local found = check_payload(p)
+    if not found then
+        -- check child payloads for E10s
+        local children = read_message("Fields[payload.childPayloads]")
+        if not children then return 0 end
+        local ok, json = pcall(cjson.decode, children)
```
trink: Could just re-use the outer scope instances

trink:
I would also avoid mixing quote styles for non-escaping purposes.

whd:
Feedback incorporated.

PR #183, opened 2016-02-01
Handle 'core' pings, update example config.

mreid-moz:
Detect and store 'core' pings. For unrecognized Telemetry ping types,
store the JSON submission in the Payload field by default.

Also update the example decoding config to use the standard field
names.

### DIFF
--- a/heka/sandbox/decoders/extract_telemetry_dimensions.lua
+++ b/heka/sandbox/decoders/extract_telemetry_dimensions.lua
@@ -230,6 +230,15 @@ local function process_json(msg, json, parsed)
         update_field(msg.Fields, "normalizedChannel" , fx.normalize_channel(auc))
 
         -- The "telemetryEnabled" flag does not apply to this type of ping.
+    elseif parsed.v then
+        -- This is a Fennec "core" ping
+        update_field(msg.Fields, "sourceVersion", tostring(parsed.v))
+        clientId = parsed.clientId
+        update_field(msg.Fields, "clientId", clientId)
+        msg.Payload = json
+    else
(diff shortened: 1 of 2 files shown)


mreid-moz:
Since this touches the dreaded `extract_telemetry_dimensions.lua`, r? @whd and @trink to make sure I don't blow up the world.

heka/sandbox/decoders/extract_telemetry_dimensions.lua
```diff
@@ -230,6 +230,15 @@ local function process_json(msg, json, parsed)
         update_field(msg.Fields, "normalizedChannel" , fx.normalize_channel(auc))
 
         -- The "telemetryEnabled" flag does not apply to this type of ping.
```
whd, line 232: This comment is to explain the absence of a following code block? (looks like bad comment spacing but probably isn't).

whd:
LGTM

heka/sandbox/decoders/extract_telemetry_dimensions.lua
mreid-moz, line 232: Yeah, it's explaining why we don't set that field for appusage pings. It's used as a dimension in the "telemetry-release" dataset, which is why it's (arguably) comment-worthy.

heka/sandbox/decoders/extract_telemetry_dimensions.lua
```diff
@@ -230,6 +230,15 @@ local function process_json(msg, json, parsed)
...
+    elseif parsed.v then
```
trink, line 233: Lets see if we can get them to fix this and make it version, but for now r+

PR #184, opened 2016-02-02
add heka plugin for sqs + s3 input

relud:
This isn't ready yet, needs testing and tests and cleanup.


relud:
this is rebased against another branch now. closing until i get it working.

PR #186, opened 2016-02-08
add sqs+s3 input for s3 buffering

relud:


heka/plugins/sqs3/sqs3_input.go
```diff
@@ -0,0 +1,178 @@
...
+    resp, err := svc.GetQueueUrl(params)
+    if err != nil { return nil, err }
+    return resp.QueueUrl, nil
+}
+
+func receive_from_queue(svc *sqs.SQS, queue_url *string) (*string, *string, *string, error) {
+    // get sqs message
+    params := &sqs.ReceiveMessageInput{
+        QueueUrl: queue_url,
+        MaxNumberOfMessages: aws.Int64(1),
```
mreid-moz, line 144: It might be good to add a `WaitTimeSeconds` setting here to put an upper bound on how long you wait for messages. I believe that it will wait for a message forever by default, which would likely cause trouble when trying to shut down the plugin if the queue is empty.

mreid-moz:
One style nit - there's a mix of camelCase and underscore_separated variable names. Go convention is camelCase, please make multi-word variable names consistent.

I don't feel super strongly about this, so if you'd rather land the code as-is, I can live with that.

heka/plugins/sqs3/sqs3_input.go
relud, line 144: It will wait for 20 seconds max, depending on how the stream is configured.

relud, line 144: Also, I need to check something, because I thought I had made that a configuration option.

relud:
camelCase - will fix

PR #190, opened 2016-02-23
Add near-realtime Firefox ADI data export

georgf:
This adds the pipeline filter & output required for [bug 1240522](https://bugzilla.mozilla.org/show_bug.cgi?id=1240522).

The idea is to track active client IDs per channel and version in hyperloglogs for a 24h window in 15min time slices, then regularly publish that data as JSON in a protected S3 bucket.
Users of the data would grab the JSON from there, e.g. via the metrics dashboard setup.

**Open issues**
- Do we need to be more clever about when to send (and consequently push to S3) new ADI JSON messages?
  - If this is something to worry about, we could presumably e.g. track when we switched to a new frame / ADI slice
  - ... or sync frame changes & timer_event()s properly, if that is possible
- I wasn't sure where `export_to_s3.lua` is supposed to live, just dropped it into `heka/sandbox/` for now.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla-services/data-pipeline/190)

### DIFF
--- a/heka/sandbox/filters/firefox_adi_rolling.lua
+++ b/heka/sandbox/filters/firefox_adi_rolling.lua
@@ -135,13 +135,32 @@ function process_message()
     return 0
 end
 
+local msg = {
+    Timestamp   = nil,
+    Logger      = "FirefoxADIRolling",
+    Type        = "FirefoxADIRolling.json",
+    Payload     = "",
+    Fields      = {
+        app         = { value = "Firefox" },
+        vendor      = { value = "Mozilla" },
+        payload_type = {
+            value = "json",
+            representation = "file-extension",
+        },
+        payload_name = { value = "Firefox ADI Rolling" },
+    },
+}
+
 function timer_event(ns)
     local now = os.time()
+
+    -- Collect the ADI data.
     local json = {
         updateTimestamp = now * 1e3,
         adi = {},
     }
 
+    -- Don't send out outdated data.
     if (now - last_update) <= ACTIVITY_WINDOW_SEC then
         local frame = get_oldest_frame(now)
         for channel,data in pairs(frame) do
@@ -152,7 +171,6 @@ function timer_event(ns)
         end
     end
 
-    -- TODO: Currently we only display this in the Heka dashboard.
-    -- This data needs to be published to S3 for consumption by other teams.
-    inject_payload("json", "Firefox ADI Rolling", cjson.encode(json))
+    msg.Payload = cjson.encode(json)
+    pcall(inject_message, msg)
 end
--- a/heka/sandbox/filters/firefox_adi_rolling.lua
+++ b/heka/sandbox/filters/firefox_adi_rolling.lua
@@ -1,0 +1,158 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+--[[
+Firefox Active Instances
+
+The memory requirements for this are roughly:
+12kb hyperloglog * framecount * 2 channels * ~50 versions
+
+Note that the version count is just a rough guess for upper limits, we will
+have to finetune this from real-world data.
+
+*Example Heka Configuration*
+
+.. code-block:: ini
+
+    [FirefoxADIRolling]
+    type = "SandboxFilter"
+    filename = "lua_filters/firefox_adi_rolling.lua"
+    message_matcher = "Logger == 'fx' && Type == 'executive_summary' && Fields[vendor] == 'Mozilla' && Fields[app] == 'Firefox'"
+    memory_limit = 300000000
+    ticker_interval = 60
+    preserve_data = true
+--]]
+require "cjson"
+require "math"
+require "os"
+require "hyperloglog"
+
+local SEC_IN_DAY = 60 * 60 * 24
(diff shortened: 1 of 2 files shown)


georgf:
Also, is the `memory_limit` for firefox_adi_rolling.lua a concern?

georgf:
@mreid-moz for review?

georgf:
There was late feedback that this does not provide the versions in the form "45.0b5" etc., so beta builds can't be told apart.
As it turns out, we don't have this data in the Telemetry data yet.
Other use-cases do a buildid to version number lookup, but i don't think this works well with Heka filters: we would have to regularly update that data from some location with some mechanism.

As Benjamin doesn't mind adding that data to the Telemetry data and it's relatively easy to do, this is the preferred path forward for me here, as this is much easier to handle in the filter.

I will have to push a follow-up to this PR once we have the client design for this.

mreid-moz:
It's a race between me and @trink to see who can get to this review first :)

heka/sandbox/filters/firefox_adi_rolling.lua
```diff
@@ -0,0 +1,176 @@
...
+
+local function clear_frame(frame)
+    for channel,data in pairs(active_instances[frame]) do
+        active_instances[frame][channel] = {}
+    end
+end
+
+local function get_frame_hll(frame, channel, version)
+    local data = active_instances[frame][channel]
+    if data[version] == nil then
```
trink, line 64: `if not data[version] then` is the Lua idiom in this case.  Also if you save it to a local you can save an additional lookup on the return.

heka/sandbox/filters/firefox_adi_rolling.lua
```diff
@@ -0,0 +1,176 @@
...
+    last_update = now
+end
+
+function process_message()
+    local ts  = read_message("Timestamp")
+    local cid = read_message("Fields[clientId]")
+    local version = read_message("Fields[version]")
+    local channel = read_message("Fields[channel]")
+
+    local function is_nonempty_string(s)
```
trink, line 125: Move this out of process_message so it isn't constantly being created/destroyed

heka/sandbox/filters/firefox_adi_rolling.lua
```diff
@@ -0,0 +1,176 @@
...
+        return type(s) == "string" and s ~= ""
+    end
+
+    if is_nonempty_string(cid) and is_nonempty_string(version) and
+       is_nonempty_string(channel) and
```
trink, line 130: You can remove the is_nonempty_string check on channel (the equality checks will suffice)

heka/sandbox/filters/firefox_adi_rolling.lua
```diff
@@ -0,0 +1,176 @@
...
+        update_activity(cid, channel, version, ts)
+    end
+
+    return 0
+end
+
+local msg = {
+    Timestamp   = nil,
+    Logger      = "FirefoxADIRolling",
+    Type        = "FirefoxADIRolling.json",
```
trink, line 141: The type field is a bit redundant; consumer of the message should just use a matcher like this
Type == "FirefoxADIRolling" && Fields[payload_type] == "json" if they want a specific output format  The plugin name can change so I probably wouldn't eliminate the Type entirely.

heka/sandbox/filters/firefox_adi_rolling.lua
```diff
@@ -0,0 +1,176 @@
...
+    Payload     = "",
+    Fields      = {
+        app         = { value = "Firefox" },
```
trink, line 144: no need to make these tables

heka/sandbox/filters/firefox_adi_rolling.lua
```diff
@@ -0,0 +1,176 @@
...
+        vendor      = { value = "Mozilla" },
+        payload_type = {
+            value = "json",
+            representation = "file-extension",
```
trink, line 148: Wow someone actually used metadata... maybe it will catch on :)

heka/sandbox/filters/firefox_adi_rolling.lua
```diff
@@ -0,0 +1,176 @@
...
+    local data = active_instances[frame][channel]
+    if data[version] == nil then
+        data[version] = hyperloglog.new()
+    end
+    return data[version]
+end
+
+local function get_oldest_frame(now)
+    local oldest_frame = (floor(now / FRAME_SIZE_SEC) + 1) % FRAME_COUNT
+    return active_instances[oldest_frame + 1]
```
trink: isn't this technically the second oldest frame

trink:
Ha, I had to r? my comments; fixed typos and removed a comment

mreid-moz:
@georgf _nudge_

georgf:
This is currently lagging due to checking whether/how this overlaps with other projects.
Once that is cleared up i can see when to re-activate it.
See also [bug 1240522](https://bugzilla.mozilla.org/show_bug.cgi?id=1240522#c23)

mreid-moz:
@georgf is this still needed?

georgf:
I dropped this in favor of other projects; i raised this as something worth picking up.
I assume the changes here at least don't apply directly to our current real-time setup anymore.

AFAIK the need still exists, but i don't know if this change is still the solution.

mreid-moz:
@georgf this seems the purview of Mission Control these days - safe to close now?

PR #193, opened 2016-03-15
Add spark job for update orphaning analysis

sapohl:
This adds the iPython notebook that runs as a scheduled Spark job on a
weekly basis on Wednesdays. The discussion of this notebook occurred in
https://bugzilla.mozilla.org/show_bug.cgi?id=1251398. It is scheduled
as:
Job 1111: update-orphaning-weekly-analysis

### DIFF
--- a/reports/update-orphaning/Update orphaning analysis using longitudinal dataset.ipynb
+++ b/reports/update-orphaning/Update orphaning analysis using longitudinal dataset.ipynb
@@ -82,24 +82,26 @@
     "today = dt.date.today()\n",
     "\n",
     "# Find the date of last Wednesday to get the proper 7 day range, and to find the last longitudinal dataset correctly.\n",
-    "last_wednesday = today\n",
-    "current_weekday = today.weekday()\n",
-    "if (current_weekday < 2):\n",
-    "    last_wednesday -= (dt.timedelta(days=5) + dt.timedelta(days=current_weekday))\n",
-    "if (current_weekday > 2):\n",
-    "    last_wednesday -= (dt.timedelta(days=current_weekday) - dt.timedelta(days=2))\n",
+    "lastWednesday = today\n",
+    "currentWeekday = today.weekday()\n",
+    "if (currentWeekday < 2):\n",
+    "    lastWednesday -= (dt.timedelta(days=5) + dt.timedelta(days=currentWeekday))\n",
+    "if (currentWeekday > 2):\n",
+    "    lastWednesday -= (dt.timedelta(days=currentWeekday) - dt.timedelta(days=2))\n",
     "\n",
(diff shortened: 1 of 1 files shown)


vitillo:
Would it possible for your job to use the latest available longitudinal dataset? See https://mail.mozilla.org/pipermail/fhr-dev/2016-March/000874.html.

sapohl:
Unfortunately, there is a dependency in our analysis script that the longitudinal dataset must have been created at least 15 days after the beginning of the date range that we want to analyze. We would rather have the job fail and not present any data in the dashboard than have the job run with a dataset that wasn't created at least 15 days after the beginning of the date range. We chose 15 days to give clients enough time to submit their telemetry data.

vitillo:
Well, you could filter fragments by submission date. Depending on the generation date of the derived dataset is very fragile as in the near future a scheduler will take care of creating the datasets and you are likely not going to be able to make assumptions on the dates any longer.

Alternatively, you could scan the list of registered tables with e.g.: `sqlContext.sql("show tables”)` and select a suitable version of the longitudinal dataset.

mreid-moz:
We could put the latest longitudinal dataset path into the metadata (assuming the job that updates the longitudinal data can write there).

mreid-moz:
@sapohl @vitillo can we file a follow-up bug for figuring out how to deal with the generation date of the longitudinal dataset in a more robust way? I'd like to land this and figure out how best to support that requirement separately.

sapohl:
I'm happy to file this bug if that's okay with @vitillo.

vitillo:
@mreid-moz we already have a way to deal with it in a robust way. The Hive metastore, which is accessible from Spark, has a table called "longitudinal" which points to the latest dataset. The metastore also has a list of all available versions of the longitudinal dataset. I am not sure why we would want to put the latest longitudinal dataset path into the metadata on S3.

mreid-moz:
@vitillo Mainly so that you could easily check the generation date against the target date.

Another thought: you could change it to a go/no-go decision by simply selecting the most recent submission date in the current data set and see if that is acceptably far past the `last_wednesday` date.

vitillo:
The datasets can be queried in Spark with `sqlContext.sql("show tables”)`. Based on that query @sapohl could get the name of the most recent longitudinal dataset that satisfies his constraints, which is what I suggested in an earlier comment.

mreid-moz:
Ah, I thought `show tables` would just list the latest table called `longitudinal`.  I see that it shows all the specific versions as well. That approach sounds good to me.

sapohl:
Thank you, @mreid-moz and @vitillo. I'll go ahead an make this change.

sapohl:
@vitillo: new version is up

vitillo:
Looks good, thanks.

PR #198, opened 2016-04-01
Fixes the d7/d30 metrics that were incorrectly calculated.

Dexterp37:
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1260715#c2


Dexterp37:
Mh, let's not merge this just yet please.

georgf:
Looking at e.g. the beta data in the backfilled data here:
https://metrics.services.mozilla.com/diagnostic-data-viewer/?dataset=fennec-v4-weekly#
... i'm confused on why `d1`/`d7`/`d30` are all higher than `actives`.

mreid-moz:
@Dexterp37 Is this ok to review/merge now?

Dexterp37:
@mreid-moz , no, we still have to go through validation once we have release data available. We already know there's a problem with the retention that needs to be fixed.

georgf:
We can still review & merge it and validate from there.
Lets just keep the bug open to keep track of this.

Dexterp37:
After a nice discussion with Gregg Lind, we cleared out a bit the definition of the d1/d7/d30 metrics. As mfinkle stated, basing our report on the profileCreation date is unreliable. I guess until we find a better solution, we have to stick with that.

georgf:
There was a background discussion on this on IRC:
https://pastebin.mozilla.org/8870431
Overall there are multiple ways to define retention and this picks one of them.
It counts a users e.g. as d7 retained if he is active on exactly the 7th day after he was first seen. This seems a bit unstable to me given users that are not active each day of the week etc.
A user active on day 6 & day 8 would not be counted as retained here.

Either way, this is a question of definition and we should clear this up separately. This seems fine, lets merge it.

georgf:
@mreid-moz, can you merge this?

PR #200, opened 2016-04-08
Add an 'activeExperimentId' field per bug 1255543

mreid-moz:


trink:
Probably want to add it here too so we don't forget it
https://github.com/mozilla-services/data-pipeline/blob/master/hindsight/input/landfill.lua#L212

PR #204, opened 2016-05-09
Extract oom crash stats daily, ultimately for use with re:dash

ddurst:
The intention is to run this script daily which will extract the previous day's data from crash-stats and write it to S3.

If this code passes muster, the script should be set up to run daily as well as back-filled from 4/27.

(I will file a separate bug to request that this data be made accessible to re:dash.)


vitillo:
`target_date` is not set properly.

vitillo:
Please remove the comment `I've never done this before (so this is one of the two crucial pieces to test), so please be merciful.`

vitillo:
Re `grouped.coalesce(1).write.parquet(s3_output)`, what's the average size of the single file generated per day?

vitillo:
Please remove commented code.

vitillo:
The Spark bits look good to me. Someone else should review the SuperSearch query as I am not familiar with it.

reports/crash_stats_oom/getting-crash-stats-for-OOM-data-to-S3.ipynb
```diff
@@ -0,0 +1,325 @@
...
+   "cell_type": "code",
+   "execution_count": 9,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "date_format = \"%Y-%m-%d\"\n",
+    "#target_date = '2016-05-02'\n",
+    "results = []\n",
```
mreid-moz: The `results` variable does not appear to be used anywhere.

ddurst:
@vitillo, the size will vary based on how much data has been recorded in socorro for the given date. The test run for 2016-05-02 produced a 1.1 MB csv. That was a few days after introduction of beta 1, so it's possible the rate was a little high; but that's conjecture as we're not sure a) if there's a lurking problem wrt OOMs and b) if OOMs would be inflated by something related to an e10s experiment.

chutten:
The query looks good to me. My only note is that you will be requesting the first page twice, which isn't the worst thing in the world, but probably can be avoided.

ddurst:
@chutten, updated (and tested to verify results are still OK). @vitillo: this should be good now.

chutten:
I'm glad you made the change. For something running daily, a duplicated page of responses could add up quickly.

vitillo:
@ddurst Please deploy the notebook on [atmo](http://analysis.telemetry.mozilla.org/) as a scheduled job. Once that is done and some data is available we can start importing it into Presto.

ddurst:
Done.

ddurst:
Updated per https://bugzilla.mozilla.org/show_bug.cgi?id=1271450#c2. Scheduled version updated as well.

vitillo:
@ddurst you have to submit a PR with the changes

ddurst:
Ah, sorry. Missed the previous merge.

PR #205, opened 2016-05-11
Use main_summary dataset to compute mau & dau.

mreid-moz:
Switch from using the executive_summary dataset to main_summary.

### DIFF
--- a/reports/engagement_ratio/MauDau.ipynb
+++ b/reports/engagement_ratio/MauDau.ipynb
@@ -22,9 +22,8 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "Unable to parse whitelist (/home/hadoop/anaconda2/lib/python2.7/site-packages/moztelemetry/bucket-whitelist.json). Assuming all histograms are acceptable.\n",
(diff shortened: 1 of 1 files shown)


reports/engagement_ratio/MauDau.ipynb
```diff
@@ -159,6 +205,52 @@
     "    diff = generated - target\n",
     "    return diff.days <= 10\n",
     "\n",
+    "\n",
+    "from moztelemetry.standards import filter_date_range, count_distinct_clientids\n",
+    "\n",
+    "# Similar to the version in standards.py, but uses subsession_start_date\n",
```
vitillo, line 211: Should we add those to `standards.py`?

mreid-moz, line 211: Yes, ideally we'd make the "mau" and "dau" functions in `standards.py` a bit more flexible so they could be used on dataframes with arbitrary column names and formats. This was the quick and dirty method.

I've filed [an issue](https://github.com/mozilla/python_moztelemetry/issues/64) in the python_moztelemetry repo to follow up with a better approach.

vitillo:
r+

PR #211, opened 2016-05-23
Skip the telemetry decode unless Logger==telemetry

mreid-moz:
This avoids creating a "telemetry-error" message for sslreports
submissions, for example. Relates to bug 1253474.

Since we don't have a message_matcher for decoders, we check
the Logger for the expected value before attempting to decode a
telemetry message.

r? @trink


trink:
I had to open up multiple source files and configs to even consider if this is correct... it looks reasonable

PR #212, opened 2016-05-27
Improve sandbox support for getting Heka data in/out of S3

trink:
For data input
- Replaces the heka-s3(list|cat) tools and the s3splitfile_input
- Supports full file compression (zst, gz, uncompressed)
- Bootstrapping support to ease configuration (for both telemetry and server logs)
  - Automatically partitions the work across the specified number of plugins
  - Dynamically creates the configurations and launches the inputs

For data output
- Replaces the s3splitfile_output
- Supports full file compression (zst, gz, uncompressed)

### DIFF
--- a/hindsight/input/heka_s3.lua
+++ b/hindsight/input/heka_s3.lua
@@ -1,0 +1,91 @@
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+-- This Source Code Form is subject to the terms of the Mozilla Public
+-- License, v. 2.0. If a copy of the MPL was not distributed with this
+-- file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+--[[
+## Reader for the S3 Heka files in compressed or uncompressed form
+
+Retrieves/reads each file from the `s3_file_list`. The primary use of this
+plugin is to feed the transformed/validated data into analysis plugins.
+
+### Sample Configuration
+```lua
+filename        = "heka_s3.lua"
+s3_bucket       = "net-mozaws-prod-us-west-2-pipeline-data"
+s3_file_list    = "files.ls.1"
+tmp_dir         = "/mnt/work/tmp"
+```
+--]]
+
+require "io"
+require "heka_stream_reader"
+require "os"
+require "string"
+
+local tmp_dir       = read_config("tmp_dir")
+local s3_bucket     = read_config("s3_bucket") or error("s3_bucket must be set")
+local logger        = read_config("Logger")
+local s3_file_list  = assert(io.open(read_config("s3_file_list")))
+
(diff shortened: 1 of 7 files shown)


hindsight/input/serverlog_s3_bootstrap.lua
```diff
@@ -0,0 +1,159 @@
...
+start_date                  = os.time({year = syear, month = smonth, day = sday})
+
+local eyear, emonth, eday   = end_date:match(DATE_FORMAT)
+end_date                    = os.time({year = eyear, month = emonth, day = eday})
+assert(end_date >= start_date, "end_date must be greater than or equal to the start_date")
+
+local num_months = (eyear * 12 + emonth) - (syear * 12 + smonth)
+
+
+local function partition_list()
```
trink, line 64: may want to pull the following three function into an input_bootstrap module

hindsight/input/heka_s3.lua
```diff
@@ -0,0 +1,91 @@
...
+
+function process_message()
+    local hsr  = heka_stream_reader.new("s3")
+
+    for fn in s3_file_list:lines() do
+        local cmd
+        local tfn = string.format("%s/%s", tmp_dir, logger)
+        local ext = fn:match("%.([^.]-)$")
+        if ext == "zst" then
+            cmd = string.format("aws s3 cp s3://%s/%s - | zstd -d -c - > %s", s3_bucket, fn, tfn)
```
mreid-moz, line 75: We should escape the S3 object name in case it contains spaces or quotes.

hindsight/input/serverlog_s3_bootstrap.lua
trink, line 64: will punt on this for now

hindsight/input/heka_s3.lua
trink, line 75: in a generic sense I would be +1 but are in full control of these objects name and if we create anything that needs escaping it is a bug (I wouldn't be opposed to adding this if we refactor out a bootstrap module)

mreid-moz, line 75: Sounds good to me.

hindsight/input/serverlog_s3_bootstrap.lua
```diff
@@ -0,0 +1,159 @@
...
+    if s3_prefix then
+        path = string.format("%s/%04d-%02d", s3_prefix, year, month)
+    else
+        path = string.format("%04d-%02d", year, month)
+    end
+
+    local cmd = string.format("aws s3 ls s3://%s/%s/ > %s", s3_bucket, path, tmp_file)
+    print(cmd)
+    local rv = os.execute(cmd)
+    if rv ~= 0 and rv ~= 256 then
```
mreid-moz, line 123: why 256?

trink, line 123: the os.execute return is shifted by 8 bits

mreid-moz:
Looks good

PR #213, opened 2016-06-06
Fix the 'hours' overflow and invalid UTF-8 data load failures

trink:


hindsight/modules/fx/ping.lua
```diff
@@ -178,7 +178,7 @@ end
 
 function hours()
     local uptime = info().subsessionLength
-    if type(uptime) ~= "number" or uptime < 0 or uptime >= 180 * SEC_IN_DAY then
+    if type(uptime) ~= "number" or uptime < SEC_IN_HOUR or uptime >= 180 * SEC_IN_DAY then
```
trink: Ha, I should have left it at 1 sec

mreid-moz: This skips any subsessions less than an hour long - change to 1sec?

PR #217, opened 2016-07-08
Break out users by version numbers for update orphaning dashboard

sapohl:
needs-review: @vitillo

Tracked in https://bugzilla.mozilla.org/show_bug.cgi?id=1284899

The dashboard has already been updated to accommodate the additional data, and to gracefully handle our old .json files that did not include data about version distribution.


reports/update-orphaning/Update orphaning analysis using longitudinal dataset.ipynb
```diff
@@ -263,7 +266,52 @@
...
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "def version_mapper(d):\n",
+    "    status, ping = d\n",
+    "    if ping.version[0] is None:\n",
+    "        return (\"-1\", ping)\n",
```
vitillo: I think `None` should work as well as key and make more sense than an arbitrary string.

sapohl:
Removed unnecessary None-check

vitillo:
lgtm

PR #231, opened 2016-11-15
Bug 1291340 - Convert & store sync data to Parquet

mreid-moz:
This change is [Reviewable](https://reviewable.io/reviews/mozilla-services/data-pipeline/231)

### DIFF
--- a/reports/sync_import/ImportSyncLogs.ipynb
+++ b/reports/sync_import/ImportSyncLogs.ipynb
@@ -1,0 +1,618 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Bug 1291340 - Import sync log data\n",
+    "\n",
+    "Read, convert, and store sync log data to Parquet form per [bug 1291340](https://bugzilla.mozilla.org/show_bug.cgi?id=1291340).\n",
+    "\n",
+    "Conversion code is ported from the [smt repo](https://github.com/dannycoates/smt)."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Unable to parse whitelist (/home/hadoop/anaconda2/lib/python2.7/site-packages/moztelemetry/histogram-whitelists.json). Assuming all histograms are acceptable.\n",
+      "Running import for 20161114\n"
+     ]
+    }
+   ],
+   "source": [
+    "from datetime import datetime as dt, timedelta, date\n",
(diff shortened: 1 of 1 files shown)


mreid-moz:
I have tested this notebook by running on an adhoc cluster, some sample data is available on S3 at
- `s3://net-mozaws-prod-us-west-2-pipeline-analysis/mreid/sync_log/v1/`
- `s3://net-mozaws-prod-us-west-2-pipeline-analysis/mreid/sync_log_device_activity/v1/`
- `s3://net-mozaws-prod-us-west-2-pipeline-analysis/mreid/sync_log_device_counts/v1/`

vitillo:
The Dataset hack isn't very future proof. Maybe adding that functionality directly to moztelemetry would make more sense? There are no tests as well and I can't help but feel that a notebook is not the right format to package this job, more so if this is going to have to run for a long while and we are planning to make changes to it.

mreid-moz:
Agreed re: the Dataset hack. Per previous discussion with @maurodoglio, I plan to send a separate PR to moztelemetry to make "summaries" generally available for filtering / manipulation, and add the ability to load records from a given set of summaries. It might still make sense to be able to use a lambda for the case when there's a very large number of summaries.

I can probably also make a change to the "standard" heka decoder to handle a missing message Payload, which would further reduce the untested code.

A notebook seemed the most approachable form for this kind of code - I'm not sure how long-lived it will be, nor who might want to make changes to it. I suspect that the pipeline's new "direct to parquet" output functionality will obsolete almost all of the non-SQL parts of this code in the next quarter or two.

mreid-moz:
> I can probably also make a change to the "standard" heka decoder to handle a missing message Payload, which would further reduce the untested code.

I took another look at this and the problem is not with the message `payload`, but rather with the assumption that field names containing a `.` contain json strings and represent nested structures inside the logical json doc. I'm not sure that's something we want to change in the normal decoder.

maurodoglio:
@mreid-moz heads up: I implemented the [changes to moztelemetry](https://github.com/mozilla/python_moztelemetry/pull/91) we talked about.

mreid-moz:
Thanks @maurodoglio !

mreid-moz:
@vitillo I filed [bug 1318762](https://bugzilla.mozilla.org/show_bug.cgi?id=1318762) to address the "overly complex notebook" issue. That will let us run python-based jobs that are more complex, but which we don't necessarily want/need to port to Scala, while still allowing for test coverage and code reuse.

Sound OK?

vitillo:
@mreid-moz OK

vitillo:
@mreid-moz, can we move this notebook to RTMO?

mreid-moz:
Yep, I'll do that next week.

mreid-moz:
Closing in favour of https://github.com/mozilla/mozilla-reports/pull/7

That PR addresses the "dataset hack" problem and moves the notebook to a better home.

PR #232, opened 2016-11-17
Bug 1315028, Bug 1307419 - Add new fields to mobile-clients dataset

fbertsch:
@maurodoglio r?

I tested this on a cluster and verified that I saw some data for each new field.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla-services/data-pipeline/232)


maurodoglio:
lgtm

PR #246, opened 2017-03-21
Added Auth token to avoid the newly-added rate limit on...

ddurst:
…the SuperSearch API. (Used this to backfill data since then, also.)

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla-services/data-pipeline/246)

### DIFF
--- a/reports/crash_stats_oom/getting-crash-stats-for-OOM-data-to-S3.ipynb
+++ b/reports/crash_stats_oom/getting-crash-stats-for-OOM-data-to-S3.ipynb
@@ -149,6 +149,7 @@
     "\n",
     "    headers = {\n",
     "        'Accept': 'application/json',\n",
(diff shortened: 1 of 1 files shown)


reports/crash_stats_oom/getting-crash-stats-for-OOM-data-to-S3.ipynb
```diff
@@ -149,6 +149,7 @@
     "\n",
     "    headers = {\n",
     "        'Accept': 'application/json',\n",
+    "        'Auth-Token': 'c61d715efd1741c3a04711d167b16304',\n",
```
mreid-moz, line 152: Is this an OK thing to put in a public repo? If so, change looks good to me.

ddurst: Fair point.

PR #247, opened 2017-03-23
Fixed atmo job getting-crash-stats-for-OOM-data-to-S3 by...

ddurst:
increasing per_page_results to 500 to handle rate limit, removed unused facet, replaced magic numbers with per_page_default. I used this to backfill data from yesterday, as if this were the regularly scheduled version.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla-services/data-pipeline/247)

### DIFF
--- a/reports/crash_stats_oom/getting-crash-stats-for-OOM-data-to-S3.ipynb
+++ b/reports/crash_stats_oom/getting-crash-stats-for-OOM-data-to-S3.ipynb
@@ -157,7 +157,6 @@
     "    param_data = {}\n",
     "    param_data['release_channel'] = \"beta\"\n",
     "    param_data['process_type'] = \"content\"\n",
-    "    # param_data['_facets'] = \"signature\"\n",
     "    param_data['_results_number'] = str(per_page_default)\n",
     "    param_data['date'] = [\">=\" + start_date, \"<\" + stop_date]\n",
     "    param_data['_columns'] = [\"date\",\n",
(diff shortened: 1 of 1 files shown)


reports/crash_stats_oom/getting-crash-stats-for-OOM-data-to-S3.ipynb
```diff
@@ -152,12 +152,12 @@
...
     "    }\n",
     "    target = 'https://crash-stats.mozilla.com/api/SuperSearch/'\n",
-    "    per_page_default = 100\n",
+    "    per_page_default = 500\n",
     "\n",
     "    param_data = {}\n",
     "    param_data['release_channel'] = \"beta\"\n",
     "    param_data['process_type'] = \"content\"\n",
-    "    param_data['_facets'] = \"signature\"\n",
+    "    # param_data['_facets'] = \"signature\"\n",
```
mreid-moz: Please remove this commented-out line (or add a comment explaining why it's there if you want to keep it).

mreid-moz (commented):
Looks good, one small change pls.

reports/crash_stats_oom/getting-crash-stats-for-OOM-data-to-S3.ipynb
ddurst: Ah, good point. I did that prior to re-run in case it impacting the data output. Those reruns appear to have worked. I'll remove it.

ddurst:
I should've measured the reruns sooner. Upping the per page results to 500 is on average in the 80s, but not always. With the limit at 100, this will also fail without a token.

PR #248, opened 2017-05-02
Updated getting-crash-stats-for-OOM-data-to-S3.ipynb to use token...

ddurst:
This ipynb has been modified in a few ways: upped the per page limit, using an Auth Token, pulling the token from S3.

@mreid-moz This *should* cover everything we've discussed, but I've been away from this for a while, so if anything should be changed about it, let me know.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla-services/data-pipeline/248)

### DIFF
--- a/reports/crash_stats_oom/getting-crash-stats-for-OOM-data-to-S3.ipynb
+++ b/reports/crash_stats_oom/getting-crash-stats-for-OOM-data-to-S3.ipynb
@@ -157,7 +157,6 @@
     "    param_data = {}\n",
     "    param_data['release_channel'] = \"beta\"\n",
     "    param_data['process_type'] = \"content\"\n",
-    "    # param_data['_facets'] = \"signature\"\n",
     "    param_data['_results_number'] = str(per_page_default)\n",
     "    param_data['date'] = [\">=\" + start_date, \"<\" + stop_date]\n",
     "    param_data['_columns'] = [\"date\",\n",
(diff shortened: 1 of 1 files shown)


reports/crash_stats_oom/getting-crash-stats-for-OOM-data-to-S3.ipynb
```diff
@@ -54,7 +55,29 @@
...
+    "    filename = \"oom.tkn\"\n",
+    "    key = \"{}/{}\".format(s3path, filename)\n",
+    "\n",
+    "    s3 = boto3.resource('s3')\n",
+    "\n",
+    "    try:\n",
+    "        obj = s3.Object(data_bucket, key)\n",
+    "    except botocore.exceptions.ClientError as e:\n",
+    "        # If the file wasn't there, that's ok. Otherwise, abort!\n",
+    "        if e.response['Error']['Code'] != \"404\":\n",
```
mreid-moz: Seems like a 404 here is not actually ok - you need this token later. Depending on how you want to handle a missing / unavailable token, it might be best to just remove the `try` block and let the code raise an error if the token can't be fetched.

mreid-moz (commented):
Looks good, one question/comment about the token-fetching logic.

mreid-moz (approved):
Looks good!

ddurst:
@mreid-moz: So.... can you merge it? Because I can't.

PR #250, opened 2017-08-31
Fixed error in date formatting causes failure if microseconds are zero.

ddurst:
It's amazing that anything works. Computers are magic.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla-services/data-pipeline/250)


mreid-moz (approved):
We should migrate this code to the [`python_mozetl` repo](https://github.com/mozilla/python_mozetl/), but for now, r+

